PR #1, opened 2026-07-05
fix(parser): closure/static-int dims in mesh/split sugar + clear DimVar/bool diagnostics (task #5)

zhen8838:
## Summary

Narrow parser fix (task #5). Two rough edges forced Qwen3-Omni kernels onto
f-string source templating and produced a misleading diagnostic; this fixes both
without adding any dynamic-layout feature.

**M1 (P1) — closure/static-int dims in mesh & split positions** (`8d85c59`)
- `_extract_dim_int` resolves a closure/global `Name` (e.g. `WARPS = 4`) to a
  static int via `_eval_ast`; rejects `bool` and non-int (`DimVar`) with a clear
  `layout dim must be a static int` diagnostic.
- Threaded `closure` through `_parse_layout_literal` / `parse_mesh_layout_sugar`
  and passed it at the `with Mesh(...)` body call site (`hir_parser`).
- Spec `docs/spec/parser.md` §1.5 updated (static-extent accepts closure/global
  int; bool/dynamic rejected).

**M2 (P2) — surface malformed layout-sugar errors** (`0ec4b28`)
- Added `LayoutSugarError(ValueError)`; the static-int violations raise it.
- `base.py` re-raises `LayoutSugarError` from the speculative sugar-parse paths
  instead of swallowing it into `_eval_static` (which turned the real "static
  int" diagnostic into `'Mesh' object has no attribute 'w'`). A genuine
  not-sugar `ValueError` still falls back as before.

## Tests

`tests/parser/hir/test_parse_shard_sugar.py`: closure mesh dims equal the
literal form, closure split-extent lock, `DimVar`/`bool` mesh rejection, `DimVar`
split-extent full-pipeline diagnostic, and split-bool single/multi-axis.

- `tests/parser/` → 124 passed.
- Broader non-GPU regression → 613 passed (one env-only `nvcc not on PATH`
  failure, passes with CUDA on PATH — unrelated to this change).

## Non-goals

No dynamic-outer tiled-layout feature; no de-templating; `DimVar` remains
rejected as a static extent (M2 only makes the error clear).

## Plan

Requirements/plan doc is kept local (the `docs/plans/` dir is gitignored per repo
convention): `docs/plans/dsl-shape-parser-fix.md`.

### DIFF
--- a/docs/spec/parser.md
+++ b/docs/spec/parser.md
@@ -149,14 +149,10 @@ axis is `Broadcast` (carries no mesh binding). A **split** extent
 participates in mesh-extent canonicalisation (factorisation), which a
 dynamic extent cannot. A dynamic split extent is rejected.
 
-A `static-extent` — whether a split extent (left of `@`) or a
-**mesh-shape** dim in a mesh's own layout tuple (e.g. `(WARPS, LANES)` in
-`Mesh(Topology("thread", THREADS), (WARPS, LANES), …)`) — may be written as an
-`integer-literal` or a `static-dim-ref`: a closure/global identifier bound to a
-static `int` (e.g. `WARPS = 4`). Both resolve identically. A `bool` is rejected
-even though it subclasses `int`, and a dynamic (`DimVar`) or otherwise non-int
-value in a `static-extent` position is rejected with a `must be a static int`
-diagnostic (never a raw AST or attribute error).
+A `static-extent` (a split extent or a mesh-shape dim) accepts an
+`integer-literal` or a `static-dim-ref` (a closure/global name bound to a static
+int); a `bool` or a dynamic (`DimVar`) / non-int value is rejected with a
+`must be a static int` diagnostic.
 
 A dynamic bare axis is admissible only where `Reshard` materialises strides
 in the **shared-engine** form — a same-storage reshard off a plain (non
--- a/docs/spec/parser.md
+++ b/docs/spec/parser.md
@@ -149,10 +149,9 @@ axis is `Broadcast` (carries no mesh binding). A **split** extent
 participates in mesh-extent canonicalisation (factorisation), which a
 dynamic extent cannot. A dynamic split extent is rejected.
 
-A `static-extent` (a split extent or a mesh-shape dim) accepts an
-`integer-literal` or a `static-dim-ref` (a closure/global name bound to a static
-int); a `bool` or a dynamic (`DimVar`) / non-int value is rejected with a
-`must be a static int` diagnostic.
+Closure/global int resolution applies to mesh-shape dims too; a `bool` or
+dynamic value in a `static-extent` is rejected with a `must be a static int`
+diagnostic.
 
 A dynamic bare axis is admissible only where `Reshard` materialises strides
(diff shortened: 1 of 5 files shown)


tests/parser/hir/test_parse_shard_sugar.py
```diff
@@ -440,3 +440,125 @@ def test_reshard_sugar_rejects_dynamic_split_axis() -> None:
...
+# and a dynamic or bool dim is rejected with a static-int diagnostic.
+
+
+def _mesh_body_fn(threads, warps, lanes):
+    """Build a nested ``@func`` whose mesh-shape sugar reads its dims from the
+    enclosing closure (``threads``/``warps``/``lanes``)."""
+
+    @func(topologies=(Topology("thread", threads),))
+    def _f(x: Tensor[(1, 128), "bf16"]) -> Tensor[(1, 128), "bf16"]:
+        with Mesh(Topology("thread", threads), (warps, lanes), ("w", "t")) as m:
```
zhen8838: Mesh should not redefine topology here. I designed it in syntax sugar before. If topology has been defined on the Function decorator, then when using it below, you only need to write the string corresponding to topology. Done. But it seems that all the current test cases and specs do not show the emphasis on this thing.

tests/parser/hir/test_parse_shard_sugar.py
```diff
@@ -440,3 +440,125 @@ def test_reshard_sugar_rejects_dynamic_split_axis() -> None:
         parse_shard_layout_sugar(
             node, lambda n: cta if n == "cta" else None, closure={"S": _S_DYN}
         )
+
```
zhen8838, line 562: The test of this file should be organized into a more unified and standardized model, such as defining a test base class. Then given the DSL function we wrote and the corresponding behavior check of its parser, now it is a separate DSL function definition function nesting and strange test naming method every time.

docs/spec/parser.md
zhen8838, line 144: Explanations have been written on the BNF grammar definition. Why do we need to write a long list of explanations here?

tests/parser/hir/test_parse_shard_sugar.py
```diff
@@ -440,3 +440,144 @@ def test_reshard_sugar_rejects_dynamic_split_axis() -> None:
...
+    def _f(x: Tensor[(1, 128), "bf16"]) -> Tensor[(1, 128), "bf16"]:
+        with Mesh(topology="thread", layout=(True, 32), names=("w", "t")) as m:
+            xr = reshard(x, (1, 128 @ (m.w, m.t)), "rmem")  # noqa: F821
+            return reshard(xr, (1, 128), "gmem")  # noqa: F821
+
+    return _f
+
+
+def _dimvar_split_extent_fn():
+    @func(topologies=(Topology("cta", 8), Topology("thread", 128)))
```
zhen8838, line 666: Here, each function of the construction test dsl func has corresponding comments to illustrate the scene of this example, and then the function name should not be underlined at the beginning, and the whole file should be changed to this mode.

docs/spec/parser.md
zhen8838, line 144: Please simplify the previous explanation here.

PR #2, opened 2026-07-05
omni-decode: settled TileFoundry compiler features (sync, dispatch, gather, async-copy, wide-load, grid-region carry)

zhen8838:
## omni-decode: settled TileFoundry compiler features

Integrates the settled compiler capabilities extracted from the Qwen3-Omni decode WIP onto a clean branch off `main`, one commit per milestone. Design/decision records live in the local plans (`docs/plans/omni-decode-integration.md`, `docs/plans/omni-decode-phase2.md`, gitignored).

### Front-end + sync
- **Grid-scope `T.sync` barrier** — a full `cta`-scope mesh lowers to a grid-wide software barrier (per-module internal-linkage counter); a partial cta slice is rejected.
- **`insert_slice(dst, update, offsets)` HIR op** — dynamic-update-slice, 1-D; the offset is a rank-0 scalar (an integer tensor for a runtime value, or a compile-time integer literal).

### Runtime-owned dispatch + ops
- **Unified reduce + sync dispatch** — codegen emits one uniform entry; the runtime derives the active per-level path (intra-warp / cross-warp / cross-cta) from the ShardLayouts via a compile-time trait. Replaces the earlier codegen-side selection. The reduce family is exposed under a single public name `tilefoundry::ops::reduce` (overloads by arity), with no scenario-coded `reduce_sharded` / `reduce_1d` names on the public surface.
- **Gather mid-axis stride fix** — a single-index gather (scalar / `(1,)`) on a non-sharded axis slices the shard layout consistently (drops or keeps the axis and remaps survivor Split axes); sharded-axis / multi-index / composed layouts pass through unchanged.
- **`async_copy` / `cp.async`** — three effect-form TIR ops (`copy_async` / `cp_async_commit` / `cp_async_wait`) forwarding to a runtime that stages each thread's fragment with 16B `__pipeline_memcpy_async` + a scalar tail.
- **Wide-load fast path** — `copy()` auto-selects a 128-bit `uint4` vector load when the source is gmem-resident and the destination is a static, contiguous, ≥128-bit fragment (verified `LDG.E.128` in SASS); it falls back to scalar otherwise, with an equivalent result.
- **Grid-region carry + in-place insert_slice + reshard-owned sync** — loop-carry lowering (single and tuple carries, the latter via `tuple_get_item`), `full_like`, `cache_update`, and an in-place `insert_slice` over the carried buffer (no replacement allocation in the loop). A cross-CTA ownership-change reshard lowers to sync-then-reshard, with the grid fence intrinsic to the reshard lowering; the intermediate and output-sink reshard paths share the same fence decision.

### Spec
The spec (`docs/spec/*.md`) is the single source of truth for the public surface: op contracts, runtime public-function contracts, and cross-function design principles (e.g. runtime-owned dispatch, cross-CTA fence ownership) all live there. Code carries a one-line purpose docstring plus local-mechanics comments only — no per-op `Spec:` back-link (the op name is the pointer). This cleanup is applied **repo-wide**, not only to the touched surface: every remaining `Spec:` back-link is removed and every long contract still in a docstring is moved to its owning spec (or, for internal tooling, reworded to local mechanics), so `scripts/spec_entropy_lint.py src` is clean across the whole Python tree. The lint is a single-directional per-changed-file pre-commit gate (long normative comment/docstring blocks → "move to spec"); it covers Python source, with C++ runtime comments relying on review. Rules in `docs/SPEC-RULES.md`.

### Verification
Every milestone is H200-validated (numeric / SASS / structural). Full suite green (`tests/ops tests/parser tests/passes` + key e2e), spec-lint clean, clang-format clean, comment-hygiene clean. Commits authored per the repo git-identity policy.

### Known boundaries (future work)
- The cross-CTA reshard owns the grid fence only; cross-CTA data redistribution (all-to-all / gather across CTAs) is not implemented and belongs to a future scenario dispatch.
- The `insert_slice` offset is a logical element start for the currently-exercised single-element (`K = 1`) window; the element-vs-tile-index contract for wider (`K > 1`) update windows is a follow-up.


docs/spec/codegen.md
```diff
@@ -103,6 +103,24 @@ target group.
   `tilefoundry::` shard tensor for a `ShardLayout` (§7) — it does not change the
   function into a `__device__`-parameter function.
 
+### 3.1 Layout-driven runtime-template selection
```
zhen8838: This content feels a little awkward here, and we need to discuss how to arrange it.

docs/spec/hir.md
```diff
@@ -204,6 +204,22 @@ input (a transposed / permuted view) MUST carry explicit strides
 from its producer; an input that is neither default-stride nor
 carries explicit strides is rejected by verify / typeinfer.
 
+**`insert_slice(dst, update, offsets)` — dynamic-update-slice.** Returns a
```
zhen8838: Later, the op of hir and tir should be annotated in the code, and referenced in the spec or automatically linked in some way. Too many details should not be added to the spec, which needs to be further discussed.

docs/spec/runtime.md
```diff
@@ -450,3 +450,66 @@ Semantics:
 Preconditions:
 
 - `tensor.layout()` is a `ShardLayout`
+
+### 3.4 `tilefoundry::ops::grid_barrier`
```
zhen8838: Why does cta's sync need to add a separate op? There was a lack of sync at the previous warp and thread level. I think it is necessary to follow the design principle of runtime and distribute it in a unified way of sync on + impl.

src/tilefoundry/codegen/cuda/tir/reduce.py
```diff
@@ -54,8 +97,13 @@ def _emit(call, ctx: CodegenContext) -> None:
...
             ws_n = ctx.name_for(call.args[2])
             wpg = int(call.target.warps_per_group)
+            # tier-2a vs tier-2b is a pure function of the operand layouts: a
+            # reduced lane axis folds within the warp (intra_cta); a reduce that
+            # crosses warps only, with each lane holding its own cells, needs the
+            # cross_warp path.
+            fn = "reduce_intra_cta" if _lane_reduced(src_ty, dst.type) else "reduce_cross_warp"
             ctx.emit(
-                f"tilefoundry::ops::reduce_intra_cta<{op_tag}, {axes_t}>"
+                f"tilefoundry::ops::{fn}<{op_tag}, {axes_t}>"
```
zhen8838: It's too unreasonable to dispatch it here. It should be dispatch inside the runtime.

tests/e2e/test_reduce_cross_warp.py
```diff
@@ -0,0 +1,54 @@
+"""End-to-end cross-warp reduce.
+
+Reducing a tensor axis whose Split maps to a warp-only mesh axis (each lane
```
zhen8838: This test should not be added here. The test of e2e should be added to the test of reduce op.

docs/spec/hir.md
```diff
@@ -80,194 +80,84 @@ structured exception that carries loop-phi-shaped SSA is
...
+Kinded pointwise unary (`neg` / `abs` / `logical_not`).
 
-Each user-callable surface name (`add` / `sub` / `cmp_eq` /
-`logical_and` / `neg` / `logical_not` / …) is registered as a
-**surface alias** ([core-ir §4](./core-ir.md)) whose `builder`
-constructs the kinded target Op directly. There are no per-name
-`Add` / `Sub` / `CmpEq` IR classes; the alias mechanism collapses 19
-sugar names into the two kinded IR Ops.
-[torch element-wise ops](https://pytorch.org/docs/stable/torch.html#pointwise-ops).
+#### Rsqrt
```
zhen8838: Isn‘t this an OP that should belong to unary?

docs/spec/hir.md
```diff
@@ -373,6 +235,15 @@ operand layout / mesh compatibility it requires and its result layout;
 there is no uniform cross-op rule imposed from outside typeinfer.
 `Reshard` is the explicit op that changes a value's layout / mesh.
 
+A single-index `Gather` — a scalar index or a rank-1 one-element (`(1,)`) index
```
zhen8838: This comment should be put in the docstring of the code.

docs/spec/runtime.md
```diff
@@ -450,3 +450,145 @@ Semantics:
...
+- `bar` points at a zero-initialized two-word counter pair reserved for this use
+
+### 3.5 Sharded reduce templates
+
+```cpp
+template <class Op, class Axes, class SrcT, class DstT, class WorkspaceT>
+__device__ void reduce_sharded(SrcT const&, DstT&, WorkspaceT&);   // entry
+template <class Op, class Axes, class SrcT, class DstT>
+__device__ void reduce(SrcT const& src, DstT& dst);                 // tier-1
+template <class Op, class Axes, class SrcT, class DstT, class WorkspaceT>
```
zhen8838: The design here is wrong. There should be only one reduce function, and then construct various reduce impl classes in it to call their operator calls.

docs/spec/runtime.md
```diff
@@ -450,3 +450,145 @@ Semantics:
...
+  lane keeps its own output cells; the lane butterfly is skipped and warps combine
+  through a workspace of `warps_per_group × 32 × n_cells` slots (one per
+  (warp, lane, cell)). Supports the SUM and MAX combines.
+
+`wpg` (`warps_per_group`) is the number of warps whose partials combine into one
+output group. The workspace is shared memory sized by the lowering (it must be
+known at allocation time). A reduction whose reduced axis crosses CTA boundaries
+is not supported (a placeholder `reduce_cross_cta` traps at compile time).
+
+### 3.6 `tilefoundry::ops::copy_async` (async gmem→smem staging)
```
zhen8838, line 507: Writing these things here, I think I confuse spec and docstring.

tests/ops/test_reduce.py
```diff
@@ -169,3 +173,128 @@ def test_reduce_max_is_signed_not_abs_max():
...
+    assert not hasattr(op, "warps_per_group")
+
+
+# ── Cross-warp reduce end-to-end (folded from the former e2e file) ───────────
+#
+# A warp-only reduction (each lane keeps its own output cell) drives the runtime
+# ``reduce_cross_warp`` path via the uniform ``reduce`` entry. Full GPU
+# compile + run + numeric compare, plus the codegen-emit shape.
+import tilefoundry  # noqa: E402
+from tilefoundry import func as _func, module as _module  # noqa: E402
```
zhen8838: My library shouldn't be imported this way; normally all imports only occur once at the very start of the file.

include/tilefoundry/runtime/cuda/runtime.cuh
```diff
@@ -322,6 +313,138 @@ CUTE_HOST_DEVICE decltype(auto) local(ShardTensor<T, GL, SL> const &st) {
         return local_impl(st, std::bool_constant<full_bc>{});
     }
 }
+
+// local_offset(st) — the projection OFFSET of this program instance's fragment
```
zhen8838: It is not feasible to implement all runtime ops in a single file; separate folders should be created based on op types for implementation.

docs/SPEC-RULES.md
```diff
@@ -1,20 +1,74 @@
 # TileFoundry Spec — Rules
 
-How `docs/spec/*.md` is written. Principle first; constraints below.
-No fixed template — pick the form that fits each construct.
+How `docs/spec/*.md` is written, and how spec and code divide responsibility.
```
zhen8838: The entire SEPC rules are overly complicated at present, and further discussions are needed to streamline them.

docs/spec/hir.md
```diff
@@ -80,239 +109,222 @@ structured exception that carries loop-phi-shaped SSA is
...
-```python
-@dataclass(frozen=True)
-class Binary(Op):
-    """ADD / SUB / MUL / DIV / FLOOR_DIV / MOD / MIN / MAX (arith);
-    EQ / NE / LT / LE / GT / GE (comparison, result dtype bool);
-    AND / OR (boolean, operands MUST be bool)."""
-    lhs: Expr
-    rhs: Expr
-    kind: BinaryKind
+Pointwise arithmetic and comparison, torch semantics with TileFoundry
```
zhen8838, line 124: The spec still needs to include operator names, parameters, fields, and their corresponding descriptions. Previously, writing out the entire dataclass directly created excessive duplication with the code implementation, so that part had to be removed. However, plain descriptive text alone is now harder to understand, so a structured format is still required to document the operator specs.
```python
@dataclass(frozen=True)
class Binary(Op):
    """ADD / SUB / MUL / DIV / FLOOR_DIV / MOD / MIN / MAX (arith);
    EQ / NE / LT / LE / GT / GE (comparison, result dtype bool);
    AND / OR (boolean, operands MUST be bool)."""
    lhs: Expr
    rhs: Expr
    kind: BinaryKind
```

docs/spec/runtime.md
```diff
@@ -406,6 +406,20 @@ Rules:
 - class-style public surfaces such as `tilefoundry::ReLUOp` are not the
   codegen-facing contract
 
+**Runtime-owned dispatch.** Where an op has more than one implementation tier
```
zhen8838, line 403: A Mermaid diagram shall be included in the runtime spec to illustrate the principles for writing ops. For instance, the op function invokes templated op implementation classes, and these classes are distributed across separate files for easier maintenance.

PR #4, opened 2026-07-13
[FEAT][hir] DSV4 decode IR contract completions

zhen8838:
## Summary

Completes the six IR contracts required by the DSV4 decode dataflow. Existing partial surfaces are extended in place — this is contract completion, not a greenfield rewrite.

- **TopK**: adds `largest` / `sorted` attributes, i64 indices, static rejection of oversized `k` and of a `Split` on the selected axis; output ShardLayout shrinks the selected axis so `size(layout) == size(shape)` holds.
- **insert_slice**: generalized from the 1-D scalar-offset form to rank-N per-axis offset tuples across typeinfer, evaluation, and lowering; the rank-N update lowers to a slice view of the existing destination buffer (no replacement allocation), gated by an end-to-end GPU oracle.
- **Pointwise kinds**: public `minimum` / `maximum` aliases; `exp` / `log` as `Unary` kinds (the first-class `Exp` op is folded into `UnaryKind.EXP`).
- **Gather**: pins the batched contract `out[b.., i.., t..] = x[b.., index[b.., i..], t..]` (`batch_dims`), keeping scalar and non-batched forms compatible.
- **Tuple returns**: `return (a, b)` and `return a, b` are a tested surface through parse, evaluation, call-site destructuring, and printer round-trip; `Tuple` is a core IR Expr (`ir.core`), matching core-ir §2.2.
- **Low-precision dtypes**: `fp8e4m3` / `f8e8m0` / `f4e2m1` as logical DType members; type inference treats them like any element type, and `Cast` is the conversion boundary (evaluator Cast for fp8e4m3 / f8e8m0; f4e2m1 is declaration-only).

Spec op entries in hir / tir / runtime use the class-form signature with a leading docstring parameter block; the format rule and canonical example live in SPEC-RULES.md § Unified Entry Format.

## Verification

- `pytest tests/` — 827 passed
- `ruff check` clean on all touched files
- `scripts/spec_rules_lint.py` on changed `docs/spec/*.md` and `docs/SPEC-RULES.md`, `scripts/spec_entropy_lint.py` on changed `src/**/*.py` — clean


docs/spec/hir.md
```diff
@@ -341,9 +341,10 @@ op that changes a value's layout / mesh.
 #### `ir/hir/math/`
 
 Pointwise arithmetic and comparison, torch semantics with TileFoundry
-type-promotion. User-callable names (`add` / `cmp_eq` / `logical_and` / …) are
-surface aliases ([core-ir §2.3](./core-ir.md)) over the kinded Ops; there are no
-per-name IR classes.
+type-promotion. Most user-callable names (`add` / `cmp_eq` / `logical_and` / …)
+are surface aliases ([core-ir §2.3](./core-ir.md)) over the kinded `Binary` /
+`Unary` Ops. A few pointwise ops are first-class per-name IR classes instead
+(e.g. `Exp`); their surface name maps to the dedicated Op, not to a kind.
```
zhen8838: It should not be modified here. Exp should still belong to a kind in unary.

docs/spec/hir.md
```diff
@@ -352,22 +353,58 @@ Binary(kind, lhs, rhs) -> Tensor    # kind: binary arithmetic, comparison, or bo
...
+  - Pointwise natural exponential `e ** x`; a first-class Op with surface name
+    `exp`, not a `Unary` kind.
 
 #### `ir/hir/tensor/`
 
 Tensor structural operations; consensus ops follow torch / numpy
 ([torch tensor manipulation ops](https://pytorch.org/docs/stable/torch.html#indexing-slicing-joining-mutating-ops)).
 
-##### Reshape / Transpose / Slice / Concat / Stack / Split / Gather / ShapeOf / Rank / Cast
+##### Reshape / Transpose / Slice / Concat / Stack / ShapeOf / Rank
```
zhen8838, line 409: The following sections have been described in sections. There is no need for this section.

docs/spec/hir.md
```diff
@@ -352,22 +353,58 @@ Binary(kind, lhs, rhs) -> Tensor    # kind: binary arithmetic, comparison, or bo
...
 Consensus torch / numpy structural ops.
 
+##### Cast
+```python
+Cast(x, dtype) -> Tensor    # x: input tensor; dtype: target element dtype
```
zhen8838: In the introduction comment of the function class here, the parameters are not newlined. It needs to be described in the following format, which should be emphasized in spec rules.
```python
"""
x: input tensor; 
dtype: target element dtype
"""
Cast(x, dtype) -> Tensor    
```

docs/spec/hir.md
```diff
@@ -392,15 +429,35 @@ Reduce(x, axes, keepdim, kind) -> Tensor    # x: input tensor; axes: reduced log
 
 ##### insert_slice
 ```python
-insert_slice(dst, update, offsets) -> Tensor    # dst: target tensor (value form returns a new tensor anchored on this buffer at lowering time); update: tensor written into the slice window; offsets: scalar runtime value or integer literal for the implemented 1-D case
+insert_slice(dst, update, offsets) -> Tensor    # dst: target tensor (value form returns a tensor anchored on this buffer at lowering time); update: tensor written into the window; offsets: per-axis window starts — a rank-0 integer scalar for a rank-1 dst, or a tuple of rank-0 integer scalars (literal or runtime), one per axis, for rank N
```
zhen8838: There are problems with the description of OP spec in the whole hir. What should be described here is the form they define, so it should be the form of python class, not the functional one.

src/tilefoundry/ir/types/dtype.py
```diff
@@ -8,9 +8,25 @@ class DType(enum.Enum):
...
     fp8e4m3 = "fp8e4m3"
+    f8e8m0 = "f8e8m0"
+    f4e2m1 = "f4e2m1"
     i32 = "i32"
     i64 = "i64"
     bool = "bool"
 
 
-__all__ = ["DType"]
+# Cast-boundary dtypes: values live only through Cast, and generic arithmetic
```
zhen8838: Delete it all here.

1. WE DON'T NEED TO ADD A COLLECTION OF LOW_PRECISION_DTYPES.

2. We should not add the typeinfer check of reject_low_precision. typeinfer is only logical, and any reasonable one should be supported.

src/tilefoundry/ir/types/__init__.py
```diff
@@ -2,7 +2,7 @@
 
 # ruff: noqa: I001 -- curated re-export order; alphabetical sort breaks staged imports.
 
-from .dtype import DType
+from .dtype import LOW_PRECISION_DTYPES, DType, reject_low_precision
```
zhen8838: delete LOW_PRECISION_DTYPES and reject_low_precision

src/tilefoundry/parser/base.py
```diff
@@ -14,6 +14,7 @@
 from tilefoundry.ir.hir.math.binary import Binary
 from tilefoundry.ir.hir.math.unary import Unary
 from tilefoundry.ir.hir.tensor.slice import Slice
+from tilefoundry.ir.hir.tensor.tuple import Tuple
```
zhen8838: why `Tuple` import from tensor.tuple ? `Tuple` should be a core IR.

tests/models/test_dsv4_ir_smoke.py
```diff
@@ -0,0 +1,85 @@
+"""DSV4 decode IR smoke.
```
zhen8838: delete this file.

tests/ops/test_insert_slice.py
```diff
@@ -2,31 +2,100 @@
...
-from tilefoundry.ir.hir.tensor.insert_slice import InsertSlice
-from tilefoundry.ir.types import DType
 from tests.ops.eval_utils import EvalCase, run_eval_case
 from tests.ops.typeinfer_utils import (
     ExpectedError,
     TypeInferCase,
     run_typeinfer_case,
     ten,
 )
+from tilefoundry.evaluator import evaluate
```
zhen8838, line 24: Internal implementations should not be exported for testing, especially _eval_rankn

tests/ops/test_insert_slice.py
```diff
@@ -416,3 +562,201 @@ def test_insert_slice_dynamic_offset() -> None:
...
+    import tilefoundry  # noqa: PLC0415
+    from tilefoundry.codegen.cuda.module import emit_cuda_module  # noqa: PLC0415
+    from tilefoundry.codegen.registry import group_functions_by_target  # noqa: PLC0415
+
+    return emit_cuda_module(
+        group_functions_by_target(tilefoundry.lower(cls, target="cuda"))["cuda"]
+    ).source
+
+
+def test_insert_slice_rankn_codegen_covers_retained_axes() -> None:
```
zhen8838: This method should not be used to test. Just look directly at the correctness of end2end.

tests/ops/test_low_precision_dtypes.py
```diff
@@ -0,0 +1,146 @@
+"""Low-precision dtype boundary.
+
+fp8e4m3 (canonical), f8e8m0 and f4e2m1 are logical dtype declarations. The
+evaluator supports Cast to fp8e4m3 and f8e8m0; f4e2m1 has no evaluator Cast.
+Generic arithmetic (Binary / Unary / MatMul / Reduce) rejects all three at
+type inference.
+"""
+from __future__ import annotations
```
zhen8838: Delete this file

1. The basic type test should appear in test ir dtype.

2. In test ops, you can add tests about different accuracy to unary and binary.

tests/ops/test_pointwise_kinds.py
```diff
@@ -0,0 +1,112 @@
+"""Pointwise surface kinds: the ``log`` unary and the ``minimum`` / ``maximum``
+binary aliases, exercised by composition oracles.
+
+Resolution is pinned explicitly: ``exp`` stays the standalone first-class
```
zhen8838: Delete this file and merge exp into unary.

tests/ops/test_topk.py
```diff
@@ -49,19 +88,147 @@ def test_topk_typeinfer(case):
...
+    x_ty = sharded((256, s), (Broadcast(),), mesh((4,)), strides=None)
+    values_ty, indices_ty = infer_call(TopK(k=6, axis=0), x_ty).fields  # select the static axis
+    assert values_ty.shape == (6, s) and indices_ty.shape == (6, s)
+    for t in (values_ty, indices_ty):
+        assert isinstance(t.layout, ShardLayout)
+        assert t.layout.layout.shape == (6, s), t.layout.layout.shape
+        assert t.layout.layout.strides == (1, 1), "concrete strides, never None"
+
+
+def test_topk_renders_in_viewer():
```
zhen8838: There is no need for render viewer in all op tests.

tests/ops/test_tuple_return.py
```diff
@@ -0,0 +1,39 @@
+"""Literal tuple return in ``@func``: the evaluator returns multiple values and
+a caller destructures a two-output helper."""
+from __future__ import annotations
+
+import torch
+
+from tilefoundry import func
```
zhen8838: Delete this file, which should be tested by parser.

docs/spec/hir.md
zhen8838: Done in ad01bbb: `exp` is now `UnaryKind.EXP` behind the alias surface and the first-class `Exp` op is deleted; the math-section intro is back to the pure-alias wording.

docs/spec/hir.md
zhen8838: Removed in ad01bbb — with exp folded into `UnaryKind` the dedicated section is gone; the Unary entry's kind list now names exp and log.

docs/spec/hir.md
zhen8838: Done in ad01bbb: the rule (class-form op signature + leading docstring block, one param per line) is pinned with a canonical example in SPEC-RULES.md § Unified Entry Format, and every callable entry in hir.md / tir.md / runtime.md is converted to it.

docs/spec/hir.md
zhen8838: Done in ad01bbb: `insert_slice` is now described as `InsertSlice(dst, update, offsets) -> Tensor`, and all hir/tir op entries use the Python-class construction form.

src/tilefoundry/ir/types/dtype.py
zhen8838: Deleted in ad01bbb: `LOW_PRECISION_DTYPES` and `reject_low_precision` are gone, the four typeinfer guards (Binary/Unary/MatMul/Reduce) are removed, and types.md drops the MUST-reject constraint — typeinfer treats the low-precision dtypes like any element type. test_unary/test_binary now assert the passthrough.

src/tilefoundry/ir/types/__init__.py
zhen8838: Deleted in ad01bbb along with the definitions in dtype.py.

src/tilefoundry/parser/base.py
zhen8838: Done in ad01bbb: `Tuple` now lives in `ir/core/expr.py` beside `Call`/`Var`/`Constant` and is exported from `tilefoundry.ir.core`; `ir/hir/tensor/tuple.py` is deleted and all import sites updated. core-ir.md §2.2 already declared it — the field name there is aligned to the implementation (`elements`).

tests/models/test_dsv4_ir_smoke.py
zhen8838: Deleted in ad01bbb.

tests/ops/test_insert_slice.py
zhen8838: Done in ad01bbb: `_eval_rankn` now builds a `@func` source string and goes through `parse_script` + `evaluate` — the hand-built Call/Function/TypeInferVisitor plumbing is gone from the eval path.

tests/ops/test_insert_slice.py
zhen8838: Done in ad01bbb: the structural walk test and both codegen source-string tests are removed; the rank-N contract is gated by the end-to-end `test_insert_slice_rankn_gpu_oracle` (non-contiguous window at a dynamic middle-axis coordinate vs a torch scatter reference).

tests/ops/test_low_precision_dtypes.py
zhen8838: Split in ad01bbb: dtype declaration basics → tests/ir_types/test_dtype.py; the Cast double-cast oracles and the f4e2m1 no-evaluator boundary → tests/ops/test_cast.py; test_unary/test_binary gain dtype-parametrized eval (f32/f16/bf16) plus low-precision typeinfer-passthrough cases. The rejection tests went away with the guard.

tests/ops/test_pointwise_kinds.py
zhen8838: Deleted in ad01bbb: exp is merged into Unary (`UnaryKind.EXP`); the composition oracle, clamp oracles, and alias-resolution checks are folded into test_unary.py / test_binary.py.

tests/ops/test_topk.py
zhen8838: Removed in ad01bbb — no viewer rendering in op tests.

tests/ops/test_tuple_return.py
zhen8838: Deleted in ad01bbb: the evaluator two-output check and the caller destructuring check moved into tests/parser/hir/test_parse_tuple_unpack.py, reusing its existing fixtures.

PR #5, opened 2026-07-14
ci: gate main with lint + full GPU pytest (self-hosted H200)

zhen8838:
## What

Adds a required GitHub Actions gate for `main`: `pre-commit run --all-files`
(lint) then the full `pytest tests/` (unit + codegen + nvcc + GPU e2e), both
running **directly inside a self-hosted runner container** on an H200 (sm_90).
No `container:` key / nested Docker — the Actions agent already runs inside the
container.

Four scoped commits:

1. `style:` tree-wide `ruff --fix` + `clang-format` so the new lint gate starts
   clean (import order / line-wrapping only; the 7 intentional DSL binding sites
   get `# noqa: F841`).
2. `feat(codegen):` parameterize the CUDA arch (was hard-coded `"80"`), threading
   `CudaTarget.arch` end-to-end (`compile.py` → `link_modules` → CMake template);
   default sm_90 (native Hopper).
3. `ci:` the workflow + `Dockerfile` + pre-commit ruff hook.
4. `build:` require `torch>=2.7` — `main` maps `DType.f8e8m0` to
   `torch.float8_e8m0fnu` (added in torch 2.7), but declared only a bare `torch`
   dep, so `import tilefoundry` fails on torch < 2.7.

## Runner requirements

The suite needs `nvcc` + a CUDA device, so it runs on a **self-hosted** runner
(hosted runners fail, not skip, the compile/e2e tests). CI expects a runner:

- labelled `tilefoundry`,
- launched pinned to one GPU (e.g. `--gpus '"device=4"'`),
- built from the included `Dockerfile` (CUDA 12.9 devel, Python 3.12,
  **torch 2.9.1+cu128**, cmake / ninja / g++ / clang-format-18, cutlass
  `include/` baked at the pinned submodule SHA at `/opt/cutlass/include`).

Jobs run their steps directly in that container; `test` symlinks the baked
cutlass headers, does an offline editable install, then runs the suite.

## Verified locally (H200, GPU 4, torch 2.9.1+cu128)

- `pre-commit run --all-files`: ruff / clang-format / spec-rules / spec-entropy — all pass.
- full `pytest tests/`: **827 passed**.

## Notes

- `pull_request` runs on the merge of this branch + base, so CI also exercises
  the current `main`.
- The `Dockerfile` uses the official `download.pytorch.org` index (portable).
- Image-rebuild automation is intentionally left out of this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,40 @@
+name: ci
+
+on:
+  pull_request:
+    branches: [main]
+  push:
+    branches: [main]
+  workflow_dispatch:
+
+concurrency:
+  group: ci-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+permissions:
+  contents: read
+
+# Jobs run directly inside the self-hosted runner container, which carries the
+# toolchain (torch, CUDA, cutlass headers at /opt/cutlass/include, ruff,
+# clang-format, pre-commit). The runner container is launched pinned to one GPU
+# (e.g. --gpus device=4), so no per-job GPU selection is needed. `tilefoundry`
+# is the runner label — set it in the runner's RUNNER_LABELS at launch.
+jobs:
+  lint:
+    runs-on: [self-hosted, tilefoundry]
+    steps:
+      - uses: actions/checkout@v4
+      - name: pre-commit run --all-files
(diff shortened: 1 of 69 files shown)


zhen8838:
Superseded by the in-repo PR (fork PRs don't trigger the new workflow since it isn't on the base default branch yet).

PR #6, opened 2026-07-14
ci: gate main with lint + full GPU pytest (self-hosted H200)

zhen8838:
## What

Adds a required GitHub Actions gate for `main`: `pre-commit run --all-files`
(lint) then the full `pytest tests/` (unit + codegen + nvcc + GPU e2e), both
running **directly inside a self-hosted runner container** on an H200 (sm_90).
No `container:` key / nested Docker — the Actions agent already runs inside the
container.

Four scoped commits:

1. `style:` tree-wide `ruff --fix` + `clang-format` so the new lint gate starts
   clean (import order / line-wrapping only; the 7 intentional DSL binding sites
   get `# noqa: F841`).
2. `feat(codegen):` parameterize the CUDA arch (was hard-coded `"80"`), threading
   `CudaTarget.arch` end-to-end (`compile.py` → `link_modules` → CMake template);
   default sm_90 (native Hopper).
3. `ci:` the workflow + `Dockerfile` + pre-commit ruff hook.
4. `build:` require `torch>=2.7` — `main` maps `DType.f8e8m0` to
   `torch.float8_e8m0fnu` (added in torch 2.7), but declared only a bare `torch`
   dep, so `import tilefoundry` fails on torch < 2.7.

## Runner requirements

The suite needs `nvcc` + a CUDA device, so it runs on a **self-hosted** runner
(hosted runners fail, not skip, the compile/e2e tests). CI expects a runner:

- labelled `tilefoundry`,
- launched pinned to one GPU (e.g. `--gpus '"device=4"'`),
- built from the included `Dockerfile` (CUDA 12.9 devel, Python 3.12,
  **torch 2.9.1+cu128**, cmake / ninja / g++ / clang-format-18, cutlass
  `include/` baked at the pinned submodule SHA at `/opt/cutlass/include`).

Jobs run their steps directly in that container; `test` symlinks the baked
cutlass headers, does an offline editable install, then runs the suite.

## Verified locally (H200, GPU 4, torch 2.9.1+cu128)

- `pre-commit run --all-files`: ruff / clang-format / spec-rules / spec-entropy — all pass.
- full `pytest tests/`: **827 passed**.

## Notes

- `pull_request` runs on the merge of this branch + base, so CI also exercises
  the current `main`.
- The `Dockerfile` uses the official `download.pytorch.org` index (portable).
- Image-rebuild automation is intentionally left out of this PR.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,7 +12,7 @@ dependencies = [
     "isl-python>=0.1.8",
     "jinja2>=3.0",
     "apache-tvm-ffi",
-    "torch",
+    "torch>=2.7",  # DType.f8e8m0 -> torch.float8_e8m0fnu, added in torch 2.7
     "graphviz>=0.20",
     "transformers>=4.57",
 ]
--- a/tests/e2e/test_mma_tir_handwritten.py
+++ b/tests/e2e/test_mma_tir_handwritten.py
@@ -16,9 +16,9 @@ import torch
 import tilefoundry
 from tilefoundry import module, prim_func
 from tilefoundry.dsl import T, Tensor
-from tilefoundry.ir.types import DType, TensorType
 from tilefoundry.ir.target.storage import StorageKind
-from tilefoundry.ir.types.shard import Mesh, Layout, Topology
+from tilefoundry.ir.types import DType, TensorType
+from tilefoundry.ir.types.shard import Layout, Mesh, Topology
 
 _OP = T.cuda.mma.SM80_16x8x16_F32BF16BF16F32_TN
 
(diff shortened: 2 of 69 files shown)


pyproject.toml
```diff
@@ -12,7 +12,7 @@ dependencies = [
     "isl-python>=0.1.8",
     "jinja2>=3.0",
     "apache-tvm-ffi",
-    "torch",
+    "torch>=2.7",  # DType.f8e8m0 -> torch.float8_e8m0fnu, added in torch 2.7
```
zhen8838, line 15: remove comment

tests/e2e/test_mma_tir_handwritten.py
```diff
@@ -32,7 +32,7 @@ def mm_device(
         c: Tensor[(16, 8), "f32"],
     ):
         atom = T.cuda.mma.atom(op=_OP)
-        with Mesh(Topology("thread", 32), Layout(shape=(4, 8), strides=(1, 4))) as warp:
+        with Mesh(Topology("thread", 32), Layout(shape=(4, 8), strides=(1, 4))) as warp:  # noqa: F841
```
zhen8838, line 35: Why do we need to use "noqa: F841" here? This should be the way I expect to write.

PR #7, opened 2026-07-15
Gather: derive shard layout instead of passing it through (fail-closed)

zhen8838:
## Symptom

`ir/hir/tensor/gather.py::_sliced_shard_layout` returned the **input** `ShardLayout` unchanged (`return sl`) in every case it could not analyze: a `Split`-axis gather, a multi-index gather, and a `strides=None` (contiguous sugar) layout. The output `TensorType.layout` then carried `size(layout.layout.shape) != size(shape)`, silently violating the `size(G) == size(T.shape)` invariant (`docs/spec/shard.md` §7.1.1). Measured on a vocab-tensor-parallel embedding lookup: `size(shape)=4096` vs `size(G)=529,530,880`.

## Root cause

`_sliced_shard_layout` (`src/tilefoundry/ir/hir/tensor/gather.py:63-113`, pre-fix) had six `return sl` sites, none of which derived a layout consistent with the gather's output shape — most notably:
- `gather.py:78` (pre-fix) — any index shape other than scalar `()` or `(1,)` passed through unconditionally, including the masked-gather shapes (`(1, 1)`, `(1, 6)`) that should derive `Partial(sum)`.
- `gather.py:82` (pre-fix) — `strides=None` (the contiguous sugar form) passed through instead of deriving prefix-product strides.
- `gather.py:91` (pre-fix) — a `Split` on the gathered axis passed through instead of deriving `Partial(sum)`.

## Fix — three derivation rules in `_sliced_shard_layout`

1. **Contiguous sugar**: `strides=None` derives concrete strides as the prefix product of the cute shape (`docs/spec/shard.md` §3 default) before any further step.
2. **Masked-gather → `Partial(sum)`**: a gathered axis carrying exactly one `Split` (and no other `Split` anywhere in `attrs`) produces `Partial(sum)` on that mesh axis, for **any** index size — each device already holds the true value at the rows it owns and a zero row elsewhere, so summing the per-device partials reconstructs the true gather.
3. **Pure slice**: otherwise, a scalar or rank-1 `(1,)` index on a non-`Split` gathered axis derives the sliced output layout (existing behavior, preserved).
4. **Fail closed**: any remaining case — multiple `Split`s anywhere in `attrs`, a composed layout, or a non-slice index combined with a `Split` elsewhere — has no derivable output layout and now raises via `ctx.error` naming the offending axis, instead of silently passing the input through.

Two pre-existing `tests/ops/test_gather.py` typeinfer cases asserted the old pass-through behavior (`sharded_axis_gather_passes_layout_through`, `multi_index_total_size_one_passes_layout_through`); these are updated to the corrected derivation (`Partial(sum)` output / fail-closed `ExpectedError`, respectively).

## Spec edit

`docs/spec/hir.md` Gather entry: replaced the stale "`batch_dims=0` retains the existing sharded slice behavior" bullet with the four rules above (`scripts/spec_rules_lint.py docs/spec/hir.md` passes).

## Test evidence

In-repo acceptance repro (`repro_g07.py`, run against this branch's editable install):
```
repro_g07.py::test_g07_vocab_tp_gather_derives_partial PASSED
repro_g07.py::test_g07_multi_row_gather_derives_partial PASSED
repro_g07.py::test_g07_sugar_slice_derives PASSED
repro_g07.py::test_g07_control_unsharded_gather_ok PASSED
4 passed
```

Full suite:
```
pytest tests/ -q
827 passed
```


src/tilefoundry/ir/hir/tensor/gather.py
```diff
@@ -60,40 +61,98 @@ def _check_batch_dims(batch_dims: int, axis: int, x_shape: tuple, idx_shape: tup
     return b
 
 
-def _sliced_shard_layout(x_ty, axis: int, idx_shape: tuple):
-    """Return the sliced ``ShardLayout`` for a pure single-index gather.
+def _prefix_product(shape: tuple) -> tuple:
```
zhen8838: This should reuse the function of our layout algebra.

src/tilefoundry/ir/hir/tensor/gather.py
```diff
@@ -60,40 +61,98 @@ def _check_batch_dims(batch_dims: int, axis: int, x_shape: tuple, idx_shape: tup
...
+        strides.append(acc)
+        acc *= int(d)
+    return tuple(strides)
 
-    A scalar index or rank-1 ``(1,)`` index on a non-sharded axis is a slice:
-    scalar indices drop that axis's cute positions, while ``(1,)`` indices keep
-    them as size-1 positions. Gather along a Split axis, multi-index gather, or
-    composed layout passes the input layout through unchanged.
+
+def _sliced_shard_layout(call, ctx, x_ty, axis: int, idx_shape: tuple):
```
zhen8838: This function is too complicated, and its principle needs to be further discussed before realizing it.

src/tilefoundry/ir/hir/tensor/gather.py
zhen8838: Reworked in commit da5ccff: layout_algebra.py's _prefix_product is now public as prefix_product (src/tilefoundry/ir/types/shard/layout_algebra.py), and gather.py's hand-rolled copy is deleted -- the gather derivation now imports and calls it directly.

src/tilefoundry/ir/hir/tensor/gather.py
zhen8838: Re-discussed and finalized as a single principle in commit da5ccff (per the updated plan): gather always produces a NEW tensor, so the output's internal cute Layout is always natural contiguous over the output shape, never inherited from the input. Only the shard attrs migrate, per mesh axis:

- Broadcast / Partial carry through unchanged (gather is a linear row selection: gather(sum_i x_i) == sum_i gather(x_i));
- a Split targeting the gathered axis becomes Partial(sum) (masked-gather rule, unchanged from v1);
- a Split targeting another axis carries through, with its logical axis renumbered for the axis removed at the gather point and idx's axes inserted in its place -- previously fail-closed, now derived;
- multiple Splits where one targets the gathered axis, or a composed layout, still fail closed.

The "Split's cute position -> logical axis" mapping is now the public shard_layout.split_target_axes, and matmul.py's _k_split_axes was migrated onto it instead of duplicating the walk. New acceptance repro (repro_g07b.py) is 3/3 green, repro_g07.py stays 4/4 green, and the full suite (827 tests) is green. See commit da5ccff for the full diff.

zhen8838:
Reworked per review — pushed commits `da5ccff` and `b2676b9` on top of this branch.

## v2: replace the case rules with a single shard-attr-migration principle

`_sliced_shard_layout` (now `_gather_shard_layout`) previously handled slicing, masked-gather, and fail-closed cases as separate hand-rolled branches, and hand-rolled its own prefix-product helper. Re-discussed and finalized as one principle:

Gather always produces a NEW tensor — the output's internal cute `Layout` is always natural contiguous over the output shape, never inherited from the input. Only the shard `attrs` migrate, per mesh axis:

- `Broadcast` / `Partial` carry through unchanged (gather is a linear row selection: `gather(Σᵢ xᵢ) == Σᵢ gather(xᵢ)`);
- a `Split` targeting the gathered axis becomes `Partial(sum)` (masked-gather rule, unchanged from v1);
- a `Split` targeting another axis carries through, with its logical axis renumbered for the axis removed at the gather point and `idx`'s axes inserted in its place — previously fail-closed in v1, now derived;
- multiple `Split`s where one targets the gathered axis, or a composed layout, still fail closed.

## Addressing the two comments

- **"too complicated" (`gather.py:74`)**: the function above replaces the old case-by-case branching with the single principle; it's now a straight per-attr map plus one fail-closed check.
- **"reuse layout algebra" (`gather.py:64`)**: `layout_algebra.py`'s `_prefix_product` is now public (`prefix_product`); `gather.py`'s hand-rolled copy is deleted. The "`Split`'s cute position → logical axis" query is now the public `shard_layout.split_target_axes`, and `matmul.py`'s `_k_split_axes` was migrated onto it instead of duplicating the walk.

## Also in this push

- Untracked `docs/plans/wi08-gather-shard-rules.md` (`b2676b9`) — `docs/plans` is gitignored by convention; the file stays on disk locally.

## Test evidence

```
repro_g07.py  : 4 passed  (v1 behavior, unchanged)
repro_g07b.py : 3 passed  (new v2 behavior)
pytest tests/ : 827 passed
```

tests/ops/test_gather.py
```diff
@@ -67,43 +67,49 @@ def test_gather_evaluate(axis, x_shape, idx):
...
     ),
-    # a non-sharded LEADING axis scalar-gather remaps the Split from axis 1 -> 0.
+    # a non-sharded LEADING axis scalar-gather renumbers the Split from axis
+    # 1 -> 0 (axis 0 removed); natural contiguous output layout.
     TypeInferCase(
         "sharded_leading_axis_scalar_remaps_split",
         Gather(axis=0),
         (sharded((6, 4, 8), (Split(1),), _M), ten((), DType.i32)),
-        sharded((4, 8), (Split(0),), _M, cute=(4, 8), strides=(8, 1)),
+        sharded((4, 8), (Split(0),), _M, cute=(4, 8), strides=(1, 4)),
```
zhen8838: remove all `cute` in tests. we only need tensor type in/out.

zhen8838: Fixed in commit 9cbc3c3: pulled the five sharded-layout cases out of the TYPEINFER_CASES table (which asserts full TensorType equality, cute layout included) into standalone tests that check only output shape and the migrated shard attrs (Split axis / Partial) -- no more cute/strides pinning anywhere in this file.

zhen8838:
Addressed the latest comment in commit `9cbc3c3`: the five sharded-layout test cases in `tests/ops/test_gather.py` no longer pin `cute=`/`strides=` — they're standalone tests now checking only output tensor shape and the migrated shard attrs (Split axis / Partial), not the internal derived cute layout.

Test evidence unchanged/still green:
```
repro_g07.py  : 4 passed
repro_g07b.py : 3 passed
pytest tests/ : 827 passed
```

PR #8, opened 2026-07-15
Reshape: divisibility-based Split refactor through adjacent-axis regroup

zhen8838:
## Summary

`Reshape`'s typeinfer rule rejected every new shape over a `Split`-carrying
`ShardLayout` with `VerifyError: "Reshape cannot express the sharded layout:
new shape does not align with the input cute factorization"`, even when the
mesh extent evenly divided the regrouped factor
(`src/tilefoundry/ir/hir/tensor/reshape.py::_carry_sharded_reshape`,
~lines 27-104). The walk that composes each new axis from cute positions had
no notion of *splitting* a single cute position across a new-axis boundary —
correct for a pure view (whole-position merges, unit-axis insertion/removal),
but it blocked two real shapes: the doc-07 head→o-group regroup
(`[1,1,64,512] Split(heads)@dev4 → [1,1,8,4096]`) and the in-graph dequant
refactor (`[8192,1024] Split(0)@dev4 → (64,128,8,128)`).

**Root cause / fix (adjacent-axis merge/split):** for a `Split`-bound cute
position of extent `N` on a mesh axis of extent `P`, subdividing `N` into an
outer factor `M` (destined for the earlier new axis) and inner residual
`m = N/M` preserves the per-device block-partition iff `P` divides `M`. The
implemented subset covers exactly this: single-`Split`-position, single
divide point. Arbitrary rank-N regroup (a device-owned block spanning more
than one split point, or multiple interacting `Split` positions) is
documented as the target contract and still fails closed — unchanged
`_carry_sharded_reshape → None → VerifyError` path.

**M4 rework (post-acceptance-review fix):** the first pass of this fix kept
the *entire* `M` sub-factor tagged `Split` whenever `P | M`, even when
`M != P`. That produces a `Split`-bound cute axis with local extent
`M/P > 1`, violating `docs/spec/shard.md` §7.1.1's invariant that
`local_shape(sl)[k] == 1` on every `Split`-bound cute dim.
`src/tilefoundry/ir/hir/sharding/reshard.py::_per_instance_strides`
(lines 52-83) relies on that invariant to assign stride `0` to `Split`-bound
axes — so an ordinary `Reshape → Reshard(rmem)` on this fix's own driving
shapes type-checked and then silently aliased distinct per-device
coordinates onto one physical slot (memory corruption) instead of the
pre-fix safe rejection. The rework factors `M` further into
`(P, Split-bound, local extent 1)` + `(M // P, plain)` whenever `M != P`,
mirroring the existing parse-time sugar canonicalization in
`src/tilefoundry/parser/sugar.py::_canonicalize_single_axis`. `TensorType.shape`
outputs are unchanged — only the internal `ShardLayout` cute factorization
gets finer.

## Test evidence

- `tests/ops/test_reshape.py`: 13/13 passing, including a migrated
  `flat_split_divides_carries` case (cute/strides updated to the finer
  factorization `(4, 8, 128)` / `(1024, 128, 1)`; `TensorType.shape`
  assertion unchanged) and a new `test_reshape_then_reshard_rmem_no_split_aliasing`
  regression guard for the `Reshape → Reshard(rmem)` aliasing defect.
- Full suite: `pytest tests/ -q` → 829 passed (net +1 test from this work).
- Spec lints: `scripts/spec_rules_lint.py` and `scripts/spec_entropy_lint.py`
  clean on all touched files.

## Spec edit

`docs/spec/hir.md`'s `Reshape` entry now has its own Unified-Entry-Format
section (previously grouped with `Transpose`/`Slice`/`Concat`/etc. under one
consensus line). It states the implemented subset (whole-position merges,
OR a `Split`-bound position dividing at a mesh-extent-dividing point,
factored into a `local_shape == 1` position plus a plain remainder per
shard.md §7.1.1) and the unimplemented target contract (arbitrary rank-N
regroup, MUST fail closed), matching the existing `Gather` `batch_dims`
precedent for documenting an intentional implementation boundary.

## Test plan

- [x] `pytest tests/ -q` — 829 passed
- [x] `pytest tests/ops/test_reshape.py -v` — 13 passed
- [x] External gap repros (`research/deepseek-v4-flash-dataflow/gap-repros/`
      in the companion TileOpsGov repo): `repro_g08.py` 3/3 passed,
      `repro_g08b.py` 2/2 passed
- [x] `scripts/spec_rules_lint.py docs/spec/hir.md` — clean
- [x] `scripts/spec_entropy_lint.py` on touched files — clean


docs/spec/hir.md
```diff
@@ -406,9 +406,49 @@ class Unary(Op):
 Tensor structural operations; consensus ops follow torch / numpy
 ([torch tensor manipulation ops](https://pytorch.org/docs/stable/torch.html#indexing-slicing-joining-mutating-ops)).
 
-##### Reshape / Transpose / Slice / Concat / Stack / ShapeOf / Rank
+##### Transpose / Slice / Concat / Stack / ShapeOf / Rank
```
zhen8838: This section should be deleted, because it has been introduced separately below.

tests/ops/test_reshape.py
```diff
@@ -89,19 +101,38 @@ def _reshape(new_shape):
...
+    # outer sub-factor (4) is exactly the mesh extent, so Split relocates
+    # there (local extent 1); the inner residual (4) merges with position 1
+    # into the trailing axis.
     TypeInferCase(
-        "straddle_fails_closed",
+        "split_divides_carries",
         _reshape((4, 32)),
         (sharded((16, 8), (Split(0),), _M),),
-        ExpectedError(match="align"),
+        sharded((4, 32), (Split(0),), _M, cute=(4, 4, 8), strides=(32, 8, 1)),
```
zhen8838: Why do you need to pass cute as a parameter? The input should be Tensor type and new shape, and what is checked is the output shape.

docs/spec/hir.md
zhen8838: Fixed in c7e10ba: removed the leftover `Transpose / Slice / Concat / Stack / ShapeOf / Rank` heading and its `Consensus torch / numpy structural ops.` one-liner (it duplicated the ir/hir/tensor/ subsection intro right above it, left over from carving Reshape into its own entry). The op names now live in that intro sentence instead; there is exactly one Reshape entry. `scripts/spec_rules_lint.py docs/spec/hir.md` passes.

tests/ops/test_reshape.py
zhen8838: Fixed in 89ef04e: the 7 sharded-carry cases (this one plus flat_split_divides_carries, merge_carries, insert_unit_axis, reshuffle_leading_unit, partial_carries, split_remaps_partial_carries) are now standalone tests asserting only the public contract — output logical shape (Tensor + new_shape in, shape out, per your point), which mesh axes stay genuinely Split/Partial, and — for the two cases where shard.md section 7.1.1's local-extent invariant matters — a semantic check via the public shard_layout_local_shape helper. No literal cute=/strides= expected-value tuples remain (git grep -n "cute=" tests/ops/test_reshape.py finds nothing). The Reshape -> Reshard(rmem) aliasing regression test is unchanged.

zhen8838:
Pushed 3 commits addressing both review comments (mechanical, no design change):

- `c7e10ba` — `docs/spec/hir.md`: removed the leftover `Transpose / Slice / Concat / Stack / ShapeOf / Rank` heading + its `Consensus torch / numpy structural ops.` one-liner, which duplicated the `ir/hir/tensor/` subsection intro directly above it (a remnant of carving `Reshape` into its own entry). The op names now live in that intro sentence; exactly one `Reshape` entry remains. `scripts/spec_rules_lint.py docs/spec/hir.md` passes.
- `89ef04e` — `tests/ops/test_reshape.py`: the 7 sharded-carry cases that previously passed `cute=`/`strides=` expected values (`merge_carries`, `insert_unit_axis`, `reshuffle_leading_unit`, `partial_carries`, `split_remaps_partial_carries`, `split_divides_carries`, `flat_split_divides_carries`) are now standalone tests asserting only the public contract: output logical shape, which mesh axes stay genuinely `Split`/`Partial`, and — for the two cases where `shard.md` §7.1.1's local-extent invariant matters — a semantic check via the public `shard_layout_local_shape` helper. `git grep -n "cute=" tests/ops/test_reshape.py` now finds nothing. The `Reshape -> Reshard(rmem)` aliasing regression test is unchanged.
- `816c4e6` — untracked `docs/plans/wi09-reshape-split-refactor.md` (plans are working-process docs, not tracked artifacts; the file stays on disk, already covered by `.gitignore`).

Verified: `tests/ops/test_reshape.py` 13/13 passed, full suite `pytest tests/ -q` 829 passed, external gap repros (`repro_g08.py` 3/3, `repro_g08b.py` 2/2) stay green, spec lints clean.

PR #9, opened 2026-07-15
relation_build: DimExpr affine inverse + opaque floordiv parameter

zhen8838:
## Summary

Gap register G-2/G-3 (TileOpsGov `11-tilefoundry-gaps-decode-module.md`):
relation-derived ops (`MatMul`, `Binary`) push an output extent through a
`DimExpr -> isl string (build) -> isl extent -> DimExpr (recover)`
round-trip. The two sides were not mutual inverses, so relation-derived ops
rejected derived shape dims (`P // 4`, `128 + P`) that hand-written shape
rules (`insert_slice`/`gather`/`topk`/`reduce`/`softmax`) already accept.

### Root cause

- `_dim_to_isl` (`src/tilefoundry/visitor_registry/relation_build.py`, the
  `DimFloorDiv` branch) raised `NotImplementedError` unconditionally — a
  `P // 4` extent never reached isl.
- `_extent_of_domain_dim` (same file) only accepted a pure constant or a
  bare coefficient-1/offset-0 `DimVar` when recovering a `ShapeDim` from
  isl's normalized `const + Sum(coeff_i * param_i)` extent. Even the
  purely-affine `128 + P` (isl normalizes to `127 + P`, then `+1`) fell
  through to a closing `ValueError`.
- `shape_from_relation` only walked top-level `DimVar` shape entries into
  its `dimvars` lookup table, so a `DimVar` nested inside a derived extent
  (the `P` inside `128 + P`) was invisible to recovery even once the affine
  reconstruction above was generalized.

### Fix

- `_extent_of_domain_dim`: generalized to rebuild the full affine sum via
  `simplify_dim(DimAdd/DimMul, ...)` over a `dimvars` table (constant term
  first, then each param term — non-unit coefficients wrap via
  `DimMul`). The single coeff-1/const-0 case stays a bit-for-bit passthrough
  so existing bare-`DimVar` recovery is unchanged.
- `shape_from_relation`: added `_collect_dimvars` to recurse into derived-dim
  operands instead of only bare top-level shape entries.
- `_dim_to_isl` / `build_domain`: `DimFloorDiv` by a positive constant now
  binds the canonicalized expression to a fresh opaque isl parameter via a
  module-level keyed table (`_floordiv_to_isl_param` /
  `_floordiv_param_name` / `_floordiv_param_expr`) — the same expression
  anywhere in the process resolves to the same parameter name. The
  parameter's bound is derived from the dividend's own value bounds via a
  new `_affine_bounds` helper (`P in [lo, hi)` => `P // c in [lo // c, (hi -
  1) // c + 1)`). The reverse mapping lets the recovery side resolve the
  opaque parameter back to its original `DimExpr` through the same path a
  bare `DimVar` uses — composite dims like `128 + P // 4` need no extra
  code.
- Out of scope, unchanged: symbolic-divisor `DimFloorDiv`, symbol x symbol
  `DimMul`, `DimMod`, `DimMax`, `DimMin` still raise `NotImplementedError`.

### Tests

- Migrates `test_build_domain_non_affine_extent_raises`, which asserted
  exactly the rejection behavior this PR replaces, into a positive test
  (`test_build_domain_floordiv_binds_opaque_param`) plus a narrower
  symbolic-divisor negative guard
  (`test_build_domain_floordiv_symbolic_divisor_raises`).
- Adds coverage for same-expression parameter dedup, affine-sum recovery,
  scaled-coefficient recovery, and floordiv round-trip.

### Spec

Updates `docs/spec/visitor-registry.md` SS4.1's `domain` description, which
said a dynamic extent becomes "an isl parameter (one per `DimVar`)" — no
longer accurate on its own now that affine combinations and the opaque
floordiv parameter are both in play.

## Test plan

- [x] `repro_g02_g03.py` (TileOpsGov gap-repros, read-only acceptance
      input) - 5/5 passed (4 flips + the bare-`DimVar` guard)
- [x] `.venv/bin/python -m pytest tests/ -q` - 832 passed
- [x] `scripts/spec_rules_lint.py docs/spec/visitor-registry.md` - clean
- [x] `scripts/spec_entropy_lint.py
      src/tilefoundry/visitor_registry/relation_build.py` - clean
- [x] `ruff check` on touched files - clean


docs/plans/wi12-relation-dimexpr.md
```diff
@@ -0,0 +1,248 @@
+---
+type: FEAT
+component: relation_build
+target_repo: tilefoundry
+---
+
+# [FEAT][relation_build] DimExpr support: affine inverse + opaque floordiv parameter
```
zhen8838: do not track plan in git

src/tilefoundry/visitor_registry/relation_build.py
```diff
@@ -22,14 +22,103 @@ def _is_const(node) -> bool:
     return isinstance(node, int) or isinstance(node, Constant)
 
 
+# Opaque isl parameter registry for ``DimFloorDiv`` extents. The same
+# canonicalized expression always binds to the same isl parameter name
+# (keyed table): ``_floordiv_param_name`` mints it on first sight,
+# ``_floordiv_param_expr`` is its reverse, consulted by the recovery side so
+# an opaque parameter resolves back to the original ``DimExpr``.
+_floordiv_param_name: dict[object, str] = {}
```
zhen8838: Such global variables should not be used for implementation. There is a problem with the whole scheme.

docs/plans/wi12-relation-dimexpr.md
zhen8838: Removed from git tracking in cef474f (`git rm --cached`) — the file stays on disk locally for reference (the docs/plans/ directory is gitignored elsewhere already; this one had been force-added). Not part of the tracked tree going forward.

src/tilefoundry/visitor_registry/relation_build.py
zhen8838: Agreed — reworked the whole scheme in 7fc5ded. Replaced the two module-level dicts with a dedicated `src/tilefoundry/utilities/isl_utility.py` module (port of nncase's `Utilities/ISLUtility.cs`):

- `to_domain(extents)` returns `(isl.set, ParamMap)` — the isl-parameter-name -> ShapeDim map is this call's own return value, never module state.
- `to_dim(pw_aff, param_map)` decodes via isl's own `ast_build.expr_from` + a generic AST visitor, not a hand-written pattern match.
- `dim_range(dim)` does interval arithmetic over every DimExpr kind, so encoding no longer special-cases per op.

`relation_build.py` is now a thin consumer; `param_map` threads through `AccessRelationResult` (extended with that field) from the two callers that need it (`matmul.py`, `binary.py`) to `shape_from_relation`. `git grep` confirms no module-level mutable dict anywhere in the diff.

This also widens acceptance beyond the v1 allowlist for free — nested floordiv, symbol×symbol mul, and mod all now derive through the same unified mechanism (repro_g02_g03b.py, 3/3 green).

zhen8838:
## wi12 v2 rework — both review comments addressed

- **Global variables / scheme rework** (7fc5ded): replaced the two module-level dicts with a dedicated `src/tilefoundry/utilities/isl_utility.py` bridge module — a port of nncase's `Utilities/ISLUtility.cs` (`dim_range` interval arithmetic, `to_domain` encode, `to_dim` decode via `ast_build.expr_from` + a generic AST visitor). `relation_build.py` is now a thin consumer; `ParamMap` threads through `AccessRelationResult` (extended with a `param_map` field) instead of any shared state.
- **Plan not tracked in git** (cef474f): `git rm --cached docs/plans/wi12-relation-dimexpr.md` — stays on disk locally, no longer in the tree.

### Test evidence
- `repro_g02_g03.py` — 5/5 green (unchanged from v1)
- `repro_g02_g03b.py` (new v2 capabilities: nested floordiv, symbol×symbol mul, mod) — 3/3 green, was 3/3 RED on the v1 branch
- `pytest tests/ -q` — 850 passed
- `git grep` — no module-level mutable dict anywhere in the diff vs `main`
- `spec_rules_lint.py` / `spec_entropy_lint.py` / `ruff check` — clean on all touched files

Commits: 7fc5ded (isl_utility rework), cef474f (untrack plan doc).

zhen8838:
## Second round of feedback addressed (41d23a9, ac12757)

1. **No nncase comments** — removed every reference to nncase / `ISLUtility.cs` from source and test comments.
2. **No new top-level directory** — moved `isl_utility.py` into `src/tilefoundry/visitor_registry/` (its only consumers already live there: `relation_build.py`, `access_relation.py`); dropped the `utilities/` packages entirely. Test file follows to `tests/analysis/test_isl_utility.py`.
3. **Fewer comments** — trimmed docstrings/comments to what the code doesn't already say for itself.
4. **`dim_range` via isl, not hand-rolled** — it now builds an isl value expression for the dim (`+`/`-`/`*`/`floor`/`mod`/`max`/`min`, arbitrarily nested) and reads the range back from isl's own `min_val`/`max_val`. The one exception: isl cannot represent a product of two non-constant terms at all (not affine, confirmed empirically — it's a genuine parser rejection, not a missing feature), so that case still computes via 4-corner interval arithmetic, recursing back through `dim_range` for its operands. Symbolic-divisor floordiv/mod is equally unrepresentable in isl and isn't exercised by any acceptance case, so it now raises `NotImplementedError` instead of carrying a hand-rolled fallback for untested territory.
5. **Dropped `ParamMap`/`DomainBuild` aliases** — `to_domain`/`to_dim` take and return plain `dict`/tuple.

Re-verified: `repro_g02_g03.py` 5/5, `repro_g02_g03b.py` 3/3, `pytest tests/ -q` 850 passed, spec/entropy lints + `ruff check` clean, no module-level mutable dict anywhere in the diff vs `main`.

PR #10, opened 2026-07-15
Elaboration-based HIR call model + exhaustive TypeInferVisitor (G-10)

zhen8838:
## Summary

Fixes gap register G-10: a wildcard (`layout=None`) `hir.Function` parameter
is documented to propagate a caller's sharded argument type through the
whole callee body, but it was silently dropped when the body ended in a
`Tuple` or `GridRegionExpr`.

**Root cause.** `TypeInferContext._compute` (`src/tilefoundry/visitor_registry/contexts.py:58-75`,
pre-fix) was an `isinstance` chain that special-cased only `Constant` /
`Var` / `Call`; every other `Expr` kind fell through to `declared =
getattr(expr, "type", None); return declared` — the node's **parse-time**
stamped type, never re-derived from its children. `_typeinfer_hir_function_call`
re-derived a callee's body through a scratch-context "overlay" that shared
IR objects across call sites, which hit this fallback for a `Tuple`-shaped
body (`repro_g10.py::test_g10_b2`) and also conflicted with "types stamped
on a node are the single source of truth" once two call sites need two
different concrete types for the same shared node.

## Fix — elaboration call model

- **`TypeInferVisitor` becomes exhaustive** (`visitor_registry/visitors.py`):
  `TypeInferContext._compute` is deleted; every `Expr` kind gets an explicit
  `visit_<Kind>` (`Var`, `Constant`, `Call` — registry dispatch, `Tuple` —
  structural over `ctx.type_of` on each element, `GridRegionExpr` — carry/body
  rule, tir's `ShapeOf` — its own declared type). An unregistered kind raises
  via `generic_visit` instead of falling back to a stamped field.
- **`elaborate(callee, arg_types, ctx)`** (`ir/hir/function.py`) replaces the
  overlay: per-call-site IR construction. A wildcard parameter binds to the
  argument's full type (incl. `ShardLayout`); an explicit `ShardLayout`
  parameter is a contract (mismatch raises). A memoized, identity-preserving
  `_Elaborator(ExprMutator)` rebuilds the body under the bound param types,
  re-stamping every changed node's type through the shared typeinfer visitor.
  Elaborating with argument types that already match the callee's current
  params returns the same `Function` instance (dedup), not a clone.
  `_typeinfer_hir_function_call` and the parser's nested-`@func`-call builder
  both route through this entry so `Call.target` is always the actual
  elaborated instance, not just `Call.type`.
- **Spec** (`docs/spec/hir.md` §1.1, `analysis.md`, `visitor-mutator.md`,
  `visitor-registry.md`) rewritten to describe the elaboration model and the
  exhaustive-visitor contract.

## Rework — 3 defects an adversarial acceptance review found

The original `repro_g10.py` didn't cover these; `repro_g10b.py` pins each one:

- **D1 — carrying `GridRegionExpr` drops a wildcard-propagated layout.**
  The parser stamps a loop's carried phi `Var`s once, at parse time
  (`parser/hir_parser.py:577`); `ir/visitor.py` correctly excludes them from
  generic mutation (they're binding sites), but `_Elaborator` never
  re-derived them either, so `TypeInferVisitor.visit_GridRegionExpr`'s direct
  read of `carried_args[0].type` stayed at the stale unsharded type even
  after the rewritten loop body carried a `Split`. Fixed: `_Elaborator`
  gained `visit_GridRegionExpr`, which re-stamps each phi from its rewritten
  `init_args` entry (hir.md §1.2: "the first-iteration value of each
  `carried_args` phi is its `init_args` entry" — the same rule the parser
  applies) before substituting the fresh phi into `body` / `yield_values`.
- **D2 — nested `Call.target` goes stale under re-elaboration.**
  `_Elaborator`'s generic rewrite rebuilt a nested Call's args and
  re-derived its `.type`, but never re-ran `elaborate()` on a Function-typed
  `.target`, so in a multi-level wildcard chain (outer → mid → leaf) the
  re-elaborated outer body's Call still pointed at the parse-time
  (unsharded) inner instance — a problem for any viewer/printer that reads
  `call.target.body`. Fixed: `_Elaborator` gained `visit_Call`, which
  re-elaborates a Function-target Call's `target` against the call's
  rewritten arg types.
- **D3 — arg-mismatch `VerifyError` lost the call-site loc.**
  The baseline reported a bind mismatch against the `Call` (so the message
  carried `at <call.loc>`); the rewritten `_bind_param_type` reported it
  against `callee` (a `Function`, whose `.loc` is always `None`), silently
  dropping the location tag. Fixed: `elaborate()` / `_bind_param_type` now
  accept an optional originating `Call`, used by `_typeinfer_hir_function_call`
  and by a loc-carrying surrogate `Call` in the parser's nested-call builder
  (which elaborates before the real `Call` exists there).

## Test evidence

- `gap-repros/repro_g10b.py` (TileOpsGov, adversarial acceptance review):
  **3/3 green** (was 3 RED before the rework)
- `gap-repros/repro_g10.py`: stays **4/4 green**
- `pytest tests/ -q`: **830 passed** (827 baseline + 3 new regression
  cases added in `tests/ir/test_function_call_typeinfer.py` and
  `tests/parser/hir/test_nested_func_call.py`; zero edits to existing
  test bodies)
- `scripts/spec_entropy_lint.py` / `scripts/spec_rules_lint.py` clean on
  touched files (one pre-existing "V1" version-stamp finding in
  `parser/base.py` predates this branch and is unrelated to this change)

Each of D1/D2/D3 was verified in isolation — reverted to a clean base,
reapplied one defect at a time, and confirmed only its own repro test
flipped green before combining.

## Test plan

- [x] `repro_g10b.py` 3/3
- [x] `repro_g10.py` 4/4
- [x] `pytest tests/ -q` full suite green
- [x] Spec lints clean on touched files


src/tilefoundry/parser/base.py
```diff
@@ -450,7 +453,23 @@ def _build_function_call(
                 f"declares {expected} parameter(s), call passed {got}"
             )
         input_args = tuple(self.expr(a) for a in node.args)
-        call = self._build_call(callee, input_args)
+        # Local import: avoids a parser <-> ir.hir.function import-order
+        # dependency at module load time.
+        from tilefoundry.ir.hir.function import elaborate  # noqa: PLC0415
```
zhen8838: Import can no longer be inserted into the function.

src/tilefoundry/ir/hir/function.py
```diff
@@ -105,54 +107,192 @@ def canonical_specialization_signature(
...
             f"callee param {param.name!r} expects {p!r}, got {arg_ty!r}",
         )
+    return p
 
 
-@register_typeinfer(Function)
-def _typeinfer_hir_function_call(call: Call, ctx) -> Type:
-    """Typeinfer handler for ``Call(target=hir.Function, args=...)``."""
-    callee: Function = call.target  # type: ignore[assignment]
+def elaborate(
```
zhen8838, line 151: I don't think it's reasonable. The function called in multiple places in IR should be the same instance.

src/tilefoundry/parser/base.py
zhen8838: Fixed in 4d4e5bf ("Move ir.hir.function <-> ir.visitor imports to module level"). Resolved the parser <-> ir.hir.function cycle structurally instead of deferring to call time: ir.hir.function's `from tilefoundry.ir.visitor import ExprMutator` and ir.visitor's `from tilefoundry.ir.hir.function import Function` are now both module-level imports, each positioned right after the symbol the *other* side needs is defined, so whichever module loads first, the back-reference finds an already-bound name instead of a partially-initialized module. Also moved the same-shaped local import of `Function` in `_resolve_function_target` (parser/base.py) to the same module-level import, since it was the identical pattern. Verified fresh-process imports of `tilefoundry`, `tilefoundry.ir.visitor`, `tilefoundry.ir.hir.function`, `tilefoundry.parser.base`, `tilefoundry.ir.hir`, `tilefoundry.script`, and `tilefoundry.dsl` all succeed independently (i.e. regardless of which side loads first), and the full test suite stays green.

src/tilefoundry/ir/hir/function.py
zhen8838: Agreed — fixed in 0ceee31 ("Same-instance elaboration: session-scoped memo keyed (callee, arg_types)"). `TypeInferContext` now carries an `elaboration_cache` dict, threaded by reference through `elaborate()`'s `TypeInferContext(module=...)` construction so it persists across one elaboration tree walk (and, via the parser's per-session `TypeInferContext`, across one parse). `elaborate()` looks it up keyed on `(id(callee), arg_types)` before doing any rebuild, and populates it once the instance is built, so two call sites of the same function with the same argument types now always resolve to the identical `Function` instance, both at parse time and under re-elaboration — no process- or module-level mutable state, matching the same-instance principle from PR #9. `docs/spec/hir.md`'s dedup wording states this session-scoped guarantee explicitly. Pinned by `gap-repros/repro_g10c.py` (2/2 green) plus in-repo regression tests in `tests/parser/hir/test_nested_func_call.py`.

zhen8838:
Pushed 3 commits addressing both review comments plus the "plans are not tracked" note:

- `4d4e5bf` — Move ir.hir.function <-> ir.visitor imports to module level. Resolved the cycle structurally (import ordering within each file), no function-local imports left in the touched files for this pattern.
- `0ceee31` — Same-instance elaboration: session-scoped memo keyed (callee, arg_types). Two call sites of the same callee with identical argument types now resolve to the identical `Function` instance, both at parse time and under re-elaboration; `docs/spec/hir.md` states the guarantee explicitly.
- `b527242` — Untrack `docs/plans/wi07-typeinfer-visitor-rebuild.md` (`git rm --cached`; file stays on disk, already covered by `.gitignore`).

Test evidence:
- `gap-repros/repro_g10c.py` (new): 2/2 green
- `gap-repros/repro_g10.py`: stays 4/4 green
- `gap-repros/repro_g10b.py`: stays 3/3 green
- `pytest tests/ -q`: 832 passed
- `spec_entropy_lint.py` / `spec_rules_lint.py` clean on touched files (same pre-existing, unrelated "V1" version-stamp finding in `parser/base.py` noted in the previous round persists, line numbers shifted only)

Both review threads replied to and marked resolved.

PR #12, opened 2026-07-16
Partial(R) pass-through gated on op-reduction commutation (G-11)

zhen8838:
## Problem
An unreduced `Partial(R)` silently crossed ops that do not commute with the carried reduction: `sigmoid(Partial(sum))` returned `Partial(sum)` (mathematically wrong — sigmoid of a sum is not the sum of sigmoids), and `Add(Partial(sum), Broadcast)` passed silently.

## Rule
A `Partial(R)` input may pass through an op IFF the op commutes with R: `op(R(x_0..x_n)) == R(op(x_0)..op(x_n))`. The judgment is (op x R)-indexed — e.g. sigmoid rejects `Partial(sum)` but passes `Partial(max)` (monotone increasing commutes with max); `Add(P,B)` rejects for sum but passes for max; `Mul(P,B)` passes for sum (linear) but rejects for max (operand sign not statically provable).

## Implementation
- Each op's typeinfer rule performs the check inline (nncase Unary precedent) — no registry, no whitelist data structure, no central check. A small pure helper `partial_reductions(layout)` extracts the carried reduction kinds.
- Error messages name the op, the offending arg, and the fix (`insert reshard(x, Broadcast) before this consumer`) — the intended reader is a DSL-writing agent iterating until the graph passes type inference.
- `docs/spec/hir.md` documents the commutation criterion; per-op tests added for every touched rule.

## Evidence
- Pinned acceptance repro: 5/5 (sigmoid/add sum-cases flip to errors; `Partial(max)` monotone pass-through, `Add(Psum,Psum)`, explicit reshard guards stay green)
- Full suite: 919 passed; spec lints clean
- Independent adversarial spot-probes of 16 (op x R) cells (tanh/add/mul/reduce sum|max|mean/matmul/transpose) all match the contract table

Do not merge — review requested.


src/tilefoundry/ir/hir/math/binary.py
```diff
@@ -21,7 +21,7 @@
 from tilefoundry.ir.core.register import register_op
 from tilefoundry.ir.core.registry import register_typeinfer
 from tilefoundry.ir.types import DType, TensorType
-from tilefoundry.ir.types.shard.shard_layout import ShardLayout
+from tilefoundry.ir.types.shard.shard_layout import ShardLayout, partial_reductions
```
zhen8838: Why import from here? It‘s incorrect.

PR #16, opened 2026-07-18
ir: add typed expression metadata

zhen8838:
## Summary

- add the frozen IRMetadata base and typed get, replace, and remove helpers
- attach immutable metadata tuples to Expr without changing equality, hashing, or repr
- reject duplicate concrete metadata classes and invalid entries with source locations

## Testing

- focused core and IR suite: 141 passed
- full suite: 926 passed in 259.67s
- Ruff passed on all changed Python files
- git diff --check passed

### DIFF
--- a/src/tilefoundry/ir/core/__init__.py
+++ b/src/tilefoundry/ir/core/__init__.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 from .errors import VerifyError
 from .expr import Call, Constant, Expr, Tuple, Var
+from .metadata import IRMetadata, get_metadata, remove_metadata, replace_metadata
 from .op import Op, ParameterInfo
 from .registry import (
     AnalysisRegistry,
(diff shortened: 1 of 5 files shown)


src/tilefoundry/ir/core/metadata.py
```diff
@@ -0,0 +1,50 @@
+from __future__ import annotations
+
+from dataclasses import dataclass, replace
+from typing import TYPE_CHECKING, TypeVar
+
+if TYPE_CHECKING:
```
zhen8838: Why do you need to use this?

src/tilefoundry/ir/core/__init__.py
```diff
@@ -4,6 +4,7 @@
 
 from .errors import VerifyError
 from .expr import Call, Constant, Expr, Tuple, Var
+from .metadata import IRMetadata, get_metadata, remove_metadata, replace_metadata
```
zhen8838, line 7: This PR lacks the modification of Spec.

zhen8838: Added in edae1e0. docs/spec/core-ir.md now owns IRMetadata, the Expr.metadata invariants and semantic exclusions, and the exact-class get, replace, and remove helper contract.

src/tilefoundry/ir/core/metadata.py
zhen8838: Removed in edae1e0. Under the merged quoted-forward-reference policy, metadata.py now uses quoted Expr annotations directly and no longer imports or guards TYPE_CHECKING.

PR #18, opened 2026-07-18
Add LayoutBase contract and nested layout printing

zhen8838:
Summary

- add a shared LayoutBase shape and domain-rank contract for Layout, ComposedLayout, and ShardLayout
- preserve nested ShardLayout composition and identity components
- print complete composed layouts canonically without flattening them

Tests

- 923 repository tests passed
- 67 focused acceptance tests passed independently
- Ruff and diff checks passed

### DIFF
--- a/src/tilefoundry/__init__.py
+++ b/src/tilefoundry/__init__.py
@@ -42,6 +42,7 @@ from tilefoundry.ir.types.shard import (
     Dynamic,
     IntTuple,
     Layout,
+    LayoutBase,
     LayoutLike,
     Mesh,
     MeshAxis,
@@ -103,7 +104,7 @@ __all__ = [
     "DType", "TensorType", "TupleType", "Type",
     "Pattern", "DimVarRangePat", "DimVar",
     # shard
-    "IntTuple", "Layout", "ComposedLayout", "LayoutLike",
+    "IntTuple", "LayoutBase", "Layout", "ComposedLayout", "LayoutLike",
     "Topology", "MeshAxis", "Mesh",
     "ShardAttr", "Split", "Partial", "Broadcast", "Dynamic", "ShardLayout",
     "S", "P", "B",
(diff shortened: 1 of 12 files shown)


src/tilefoundry/__init__.py
```diff
@@ -103,7 +104,7 @@ def view(root, *, port: int = 0, open_browser: bool = True) -> int:
     "DType", "TensorType", "TupleType", "Type",
     "Pattern", "DimVarRangePat", "DimVar",
     # shard
-    "IntTuple", "Layout", "ComposedLayout", "LayoutLike",
+    "IntTuple", "LayoutBase", "Layout", "ComposedLayout", "LayoutLike",
```
zhen8838: What is the difference between Layout base and Layout like?

zhen8838: Removed `LayoutLike` entirely. `LayoutBase` is now the sole layout-family boundary across `ComposedLayout`, `ShardLayout`, `TensorType` annotations, public exports, tests, and specs.

src/tilefoundry/ir/types/shard/layout.py
```diff
@@ -1,19 +1,31 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import Optional, Union
+from typing import Optional
+
+
+def _domain_rank(shape: tuple) -> int:
```
zhen8838: should move into layout algebra or reuse exists function

zhen8838: Fixed in fdf72d1. LayoutBase.domain_rank now reuses the shard int-tuple flattening helper instead of defining layout-local algebra. The helper now follows the CuTe tuple rule (tuples recurse; every non-tuple is a leaf), so symbolic and launch-provided dimensions remain countable. Added coverage with a nested shape containing None.

PR #19, opened 2026-07-18
schedule: add public stage dispatch contract

zhen8838:
## Summary

- add the thin public Schedule protocol, options, result, error, and explicit stage solve facade
- serialize the stable ScheduleReport summary as sorted JSON and Markdown
- cover exact fake-owner lookup, validation, no fallback, and report output

## Tests

- 48 focused and core tests passed
- Ruff passed for the new package and tests

## Integration boundary

This slice defines and tests the public contract only. Real Target.service integration and CTA service registration remain intentionally owned by S06.


src/tilefoundry/schedule/__init__.py
```diff
@@ -0,0 +1,21 @@
+"""Public scheduling API."""
+
+from __future__ import annotations
+
+from tilefoundry.schedule.api import (
```
zhen8838: There is no need to add schedule.api . It is more reasonable to export Schedule directly.

src/tilefoundry/schedule/api.py
```diff
@@ -0,0 +1,84 @@
...
+
+    module: Module
+    report: ScheduleReport
+
+
+class ScheduleError(Exception):
+    """An actionable scheduling failure exposed to callers."""
+
+
+def solve(
```
zhen8838: There is no need to package a function separately for such a simple function.

zhen8838: Removed the free solve wrapper and ScheduleError. Callers now obtain the exact-stage service from the root Target and invoke service.solve(...) directly.

src/tilefoundry/schedule/__init__.py
zhen8838: Removed schedule.api and report.py. Schedule, ScheduleOptions, ScheduleResult, and ScheduleReport now live directly in schedule/__init__.py.

PR #32, opened 2026-07-27
refactor: give execution context and hardware facts one owner each

zhen8838:
Three constructs each had their facts written down in more than one place, and
nothing forced the copies to agree. This gives each one a single owner.

- **Where a program runs** was declared on an authored HIR `Function`, while the
  `Module` that owned it also carried a target. Now the Module declares it and a
  Function resolves it lexically.
- **What the hardware is** existed twice: as attributed TOML facts *and* as
  Python dataclass defaults. Now the installed documents are the only place a
  hardware number is written.
- **How an algorithm reads hardware** had no boundary at all — a consumer
  reached into `target.device.<field>` and branched on the concrete target. Now
  it declares the aggregate it wants and the Target package registers the
  conversion.

Baseline `915 passed, 4 failed` → **`976 passed, 4 skipped, 0 failed`**. The four
skips are the minicpm3 oracle tests, which import `transformers.models.minicpm3`;
transformers 5.x does not ship it, so they cannot pass on a conforming install of
the declared dependency. They were failing before and are now skipped with that
reason stated.

## Execution context belongs to the Module, not the Function

A `Function` no longer says where it runs. The execution context — a Target plus
an ordered Topology hierarchy — is declared by the owning Module and resolved
over the owner chain, so a reusable child inherits it instead of restating it.

`topologies=None` means inherit and `()` means explicitly topology-free; the two
must stay distinguishable, because "I did not say" and "I said none" describe
different programs. Only the outermost Module declares a Target, and a child
declaring its own is a construction error rather than a silent override.

A Module also owns its subtree, which is what makes the resolution stable:

```python
on_cuda = Module("root", (), "forward", modules=(child,), target=CudaTarget())
on_cpu  = replace(on_cuda, name="copy", target=CpuTarget())

on_cuda.modules[0].resolve_target()  # CudaTarget
on_cpu.modules[0].resolve_target()   # CpuTarget
```

Without that, copying a root and retargeting the copy would silently retarget
the original's children through the shared back-link.

### The authoring surface keeps both forms

`@func(target=..., topologies=...)` still works: declaring either binds the name
to an implicit single-function Module carrying that context. A plain `@func`
still yields a `Function`, which is what lets a `@module` class body detect its
children by type.

The topology hierarchy is a `topologies` assignment at the top of the class body
rather than a decorator argument, because a class body's decorators run *after*
its preceding assignments bind:

```python
@module(entry="attend")
class GqaOnline:
    topologies = (Topology("cta", NUM_CTA),)

    @func
    def attend(q: Tensor[..., Mesh(topology="cta")], ...): ...
```

`canonical_shard_layout` consumes `mesh.shape` at parse time to build the
`ShardLayout`, so the hierarchy has to be resolvable while the body is being
parsed. A decorator argument is evaluated before the body runs and cannot serve
that lookup; a class attribute can.

## Hardware facts live in installed documents

The two combined device-and-architecture TOMLs are replaced by four independent
documents, each complete in its own right. A target is the pair composed through
a declared compatibility, never a single record. The value classes now declare
shape, not content — none carries a default, so none can be constructed without
a document behind it.

```toml
[spec]
schema = "tilefoundry.cuda.device/v1"
kind = "device"
id = "nvidia.h200_sxm"

[compatibility]
architectures = ["nvidia.sm90"]

[facts.memory.hbm.bandwidth]
value = 4800000000000
unit = "byte/s"
origin = "vendor"
source = "https://www.nvidia.com/en-us/data-center/h200/"
conditions = "4.8 TB/s peak HBM3e bandwidth, decimal"

[facts.memory.l2.bandwidth]
status = "unavailable"
conditions = "NVIDIA publishes no static L2 bandwidth to use for a roofline."
```

An unavailable fact omits `value` entirely instead of storing the string
`"unavailable"` where a number belongs, so no caller can read a placeholder as a
quantity. `origin` separates a published figure from one measured on the
described host, a citation, a derivation, and a reading no source states — the
AMX unit count is `estimated` because it was read off a thread-scaling curve,
and calling that `measured` would overstate it.

Selecting a resource is not the same as editing one. `CudaTarget(architecture=
"nvidia.sm90")` names another complete, attributed document; supplying a number
with no document behind it does not, and is not admitted. A resolved target
retains each document's ID and content digest, so a compiled artifact can name
the exact resources it was built against.

Compiler policy leaves the hardware documents: a fixed-wave parallel capacity is
a scheduling input even when its current value happens to equal a device count.

### The split is by what a fact is a property of

Per-SM limits — resident CTAs, shared memory per SM and per CTA, register file
per SM — moved from the H200 device to SM90, and the AMX X/Y/Z register files
from the M2 Pro package to AppleAmx. Every product built on a microarchitecture
shares its limits, so recording them per-product invites drift. The existing code
already knew this; `amx/atoms.py` carried the comment *"The X/Y/Z register files
are AMX ISA geometry, not a per-part figure"* while reading them off the device.

## Algorithms project facts through one boundary

`Target.as_facts(FactsType, query=None)` converts a specification into the
immutable aggregate a requesting algorithm declares. Lookup is by the exact
`(Target concrete type, Facts type)` pair — a base-class registration does not
serve a subclass, because two targets sharing a base can describe different
hardware and inheriting a conversion would project facts for the wrong device.

A missing conversion fails immediately rather than falling back to a default, and
a duplicate registration fails rather than replacing, so a projection cannot
depend on import order. The query belongs to the requesting algorithm: a
hardware-only projection requires its absence, a program-dependent one validates
its own private type, and no common query base is introduced.

The registry names no concrete target, so adding a backend adds a registration
rather than a branch. The aggregates exercised here are local to the tests: a
production algorithm declares its own, so this adds the boundary without fixing a
shape no caller has asked for yet.

## Two silent failures worth recording

Both would have produced wrong answers rather than errors, which is why they
are recorded here rather than just fixed:

- Moving `shared_memory_per_cta_bytes` to the architecture left the CTA analysis
  service reading it off the device. The read used `getattr(..., None)`, so it
  returned `None` instead of raising. Eight atom-selection tests were the only
  thing between that and a runtime surprise; nothing in the moved code failed.
- Putting the resolved document IDs on the target dataclass let them into
  `__eq__`. Codegen groups CUDA functions by comparing targets, so two targets
  with **identical facts** but different provenance would have been reported as
  *"differing Target facts"*. Identity records where a value came from, not what
  it says, so it stays out of equality — as `_services` already did.

The printer had a related gap: it decided defaultness from the architecture
alone, so a custom device printed as a bare `CudaTarget()` and was lost. It now
emits each side independently — an installed resource as its ID, a directly
supplied one as the constructor that rebuilds it, with the imports that
constructor needs — and the round-trip tests execute the emitted source rather
than string-matching it.

## Specs

Twelve spec documents are updated in the same change: `analysis`,
`architecture`, `cli`, `code-organization`, `core-ir`, `hir`, `inspection`,
`parser`, `passes`, `runtime`, `schedule`, and `target`. Every removed symbol was grepped across
`docs/spec/`, `src/`, and `tests/` — one spec still carried a worked example
calling the deleted `root.target.service(...)`.

Three pre-existing spec drifts were corrected while in the same sections:
`AppleM2Pro.l1_capacity_bytes` and `l2_bandwidth_bytes_per_second` were specified
but never existed, `throughput_for` really takes `(unit, dtype)`, and
`topology_limit("cta")` really returns `None` rather than `sm_count`.


src/tilefoundry/inspection/python_printer.py
```diff
@@ -978,27 +978,92 @@ def _target_str(target: Target) -> str:
...
+                ("architecture", _architecture_arg(target)),
+                ("device", _device_arg(target)),
+            )
+            if value is not None
+        ]
+        return f"CudaTarget({', '.join(arguments)})"
     raise TypeError(f"unsupported target for Python printing: {type(target).__name__}")
 
 
+def _architecture_arg(target: CudaTarget) -> str | None:
```
zhen8838, line 993: Why is this kind of target-related printing placed in public places?

PR #57, opened 2026-08-04
docs(readme,examples,scripts): make the Quick Start a prompt that has been run

zhen8838:
## Why

- `README.md` is twelve lines, one of which says the project "is not yet ready for
  use", while `tilefoundry` 0.0.1 has been on PyPI since 2026-08-02 and
  `readme = "README.md"` makes this file the PyPI landing page as well as the
  GitHub front page.
- A Quick Start for an agentic platform cannot be a hand-authored `@module`
  snippet: the premise is that you do not write the description or the kernels, so
  a snippet advertises the opposite of the product. It has to be a prompt.
- Publishing a prompt that had never been run would be a claim, not a Quick Start.
  The prompt in this README produced the example in this branch, unedited.

## What

- `README.md` rewritten: banner, badges and Documentation / Installation /
  Examples centred under a rule, one sentence of what this is, Latest News, then
  Installation and Quick Start as separate sections, then License. The Quick Start
  is thirteen lines of prompt and closes on the measured outcome. Every link is
  absolute, because relative paths 404 on PyPI; the banner is a GitHub attachment
  rather than a committed file, since it is 5.4 MB and git history cannot be made
  to forget it later.
- `examples/qwen3_1_7b-tilelang/` added, 13 files. Shipped the way the other three
  are -- reference (`ref_src/`), twin (`fast/twin.py`), kernels
  (`fast/kernels.py`), entry point (`run.py`) -- with the workbench the work was
  done with left out: no verify.sh driver, three profilers, two diagnostics,
  activation dumper or engine test. `fast/test_kernels.py` stays because it is the
  torch spelling of every kernel and the siblings ship that as `torch_ref.py`;
  `fast/arbitrate.py` stays because the f64 arbitration has no other copy. Its
  README carries the same four sections as the others, with the agent's own prose
  intact underneath.
- `scripts/no_machine_paths_lint.py`: a `-` immediately before the slash may now
  precede a match. `:-` put one exactly where a path begins, so `${VAR:-<a path>}`
  was the one shape the checker could not see -- it passed this example while the
  example still carried a checkpoint path that way. `--option=<a home directory>`
  was invisible for the same reason. Both forms added to the caught list in
  `tests/scripts/test_no_machine_paths_lint.py`.
- **No model description is changed.** An earlier revision of this branch carried
  the example's attention-scale finding back into all seven; see Risk.

## Contract

- No public API, IR, CLI surface or spec text changes. `src/` is untouched.
- `no-machine-paths` gets stricter, not looser: two shapes that used to pass are
  now reported. The repository is clean under the new pattern, so no existing file
  needs an allow marker.
- Nothing in CI runs anything under `examples/`, before or after. Each example
  states its own "verified at v0.0.1, not verified since" at the top; the new one
  now does too.

## Verification

- `pytest tests/scripts -q -n 4` — **22 passed in 19.57s**, exit 0
  (`test_results/pr57-checker.txt`). The 2 new cases are the shapes that leaked.
- `pre-commit run --files README.md scripts/no_machine_paths_lint.py tests/scripts/test_no_machine_paths_lint.py`
  — all hooks Passed, exit 0.
- `git ls-files | xargs grep -nIE '/data[0-9]*/|/home/[a-z][a-z0-9._-]*/|/Users/[A-Za-z]'`
  — **0 files**, the checker's own source aside. Grepped directly rather than
  trusting the checker, which is how the leak was found.
- CI on the identical tree, run 30905142681: `lint` pass 17s, `test` pass 5m3s,
  `install-smoke` pass 4m20s.
- The example's own numbers are the agent's, re-stated not re-run: 612.5 tok/s on
  one H200 over 2048 new tokens against 14.8 tok/s through the evaluator;
  `tilefoundry check` passing at context extents 0/1/255/1024 and on all 57 outputs
  of the whole model in one step; 255/256 teacher-forced positions agreeing with
  Hugging Face.
- The claim that the descriptions must not change is measured, not assumed:
  `qwen3_1_7b` authored HIR through the Evaluator, both scale orderings, at context
  0 / 24 / 255 -- same output to every digit reported, against Hugging Face in bf16
  and against an f64 arbiter.

## Risk

- The example is a snapshot, not a supported path. It pins `tilelang` 0.1.12,
  `torch` 2.9.1+cu128, CUDA 12.8 and one H200 in its §1, and nothing re-runs it, so
  it will rot silently. That is the same risk the other three already carry.
- `README.md` names a measured throughput. If the example stops reproducing it,
  the front page is wrong and nothing will say so.
- The Quick Start's honest cost -- hours of agent time and a GPU -- is stated, but
  the prompt has been demonstrated on exactly one model, from the catalog, on one
  machine. It does not yet support the claim "your own model too".
- The attention-scale finding is real and is **not** acted on here. It was
  arbitrated on the tilelang twin, where the multiply really is bf16, and does not
  transfer to authored HIR. Where it did change something it made it worse:
  `kimi_linear_48b_a3b`, NoPE at context 24, scaling `q` sits 0.00408 from f64 and
  scaling the score 0.00462, with Hugging Face itself at 0.00454 -- the original
  ordering was the closer of the two, and the change pushed the uniform component
  contract from 2.75 to 3.25 roundings and failed `test_mla_nope_matches_hf` on 2 of
  2304 elements. Carrying it into the descriptions needs its own per-model
  arbitration.
- Separately worth tracking: that test now has only 2.75 of its 3 permitted
  roundings to spare while Hugging Face itself sits 2.3 roundings from f64. Two
  independent errors of that size will reach the bound eventually, with or without
  this branch. #56 removes the test; if it survives, the bound wants a look.
- `Documentation` in the header points at the published site. It resolves today
  (200), but nothing in this repository builds or checks it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,73 @@
-# TileFoundry
+<p align="center">
+  <img src="https://github.com/user-attachments/assets/3e29ee3e-2fae-4243-ba73-0efc04ac7645" alt="TileFoundry" width="100%">
+</p>
 
-[![License: MIT](https://img.shields.io/badge/License-MIT-blue.svg)](LICENSE) ![Status](https://img.shields.io/badge/status-early%20development-orange)
+---
+
+<p align="center">
+  <a href="https://pypi.org/project/tilefoundry/"><img src="https://img.shields.io/pypi/v/tilefoundry.svg" alt="PyPI"></a>
+  <img src="https://img.shields.io/badge/status-early%20development-orange" alt="Status: early development">
+  <a href="https://github.com/tile-ai/TileFoundry/blob/main/LICENSE"><img src="https://img.shields.io/badge/License-MIT-blue.svg" alt="License: MIT"></a>
+</p>
+
+<p align="center">
+  <a href="https://tile-ai.github.io/TileFoundry.github.io/">Documentation</a> &middot;
+  <a href="https://github.com/tile-ai/TileFoundry#installation">Installation</a> &middot;
+  <a href="https://github.com/tile-ai/TileFoundry/tree/main/examples">Examples</a>
+</p>
 
 **TileFoundry** is a tile-based, agentic platform for automatic high-performance program generation across hardware.
 
-> [!NOTE]
-> TileFoundry is in an early design and development stage. APIs and architecture are still evolving, and the project is not yet ready for use.
+## Latest News
+
+- 08/2026 🎉: **TileFoundry 0.0.1 is on PyPI** — the first public release.
+- 08/2026 📦: Four [worked examples](https://github.com/tile-ai/TileFoundry/tree/main/examples) added — Qwen3-1.7B (tilelang), Qwen3.5-35B-A3B (tilelang), MiniCPM3-4B (CuTeDSL) and granite-4.0-h-small (CUDA C) — each one a real agent run kept whole, with the decode throughput it measured.
+
+## Installation
+
+TileFoundry needs Python 3.12 or newer.
+
+```sh
+pip install tilefoundry
+```
+
+Check the install — it prints the commands an agent will ask:
+
+```sh
+tilefoundry
+```
+
+Running a model an agent generates additionally needs one NVIDIA GPU and the
+checkpoint already on disk.
+
+## Quick Start
+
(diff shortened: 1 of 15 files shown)


zhen8838:
banner_2

PR #64, opened 2026-08-07
fix(passes): carry the Module's Target into the lowered TIR

### LINKED ISSUE #63: HirToTirPass discards the Module's Target, so a Module can be compiled for the wrong backend
# HirToTirPass discards the Module's Target, so a Module can be compiled for the wrong backend

**Affects:** `main` @ `e55e4b0` (reproduced on the current tip)

## Summary

`HirToTirPass` constructs every `PrimFunction` without passing `target=`, so the
Target declared by the owning `Module` never reaches the lowered TIR.
`PrimFunction.target`'s `field(default_factory=default_target)` (cuda / sm_90 /
h200) silently takes over.

The consequence is not limited to a wrong SM arch. `group_functions_by_target`
keys on `fn.target.name`, so a Module declaring a non-CUDA backend is handed to
the **CUDA** emitter.

## Reproduction

Only the Module's `target=` varies; the function, params and body are identical.

```python
from tilefoundry.codegen.registry import group_functions_by_target
from tilefoundry.ir.core import Var
from tilefoundry.ir.core.module import Module
from tilefoundry.ir.hir.function import Function as HirFunction
from tilefoundry.ir.types import DType, TensorType
from tilefoundry.passes.transforms import HirToTirPass
from tilefoundry.target.amx.target import AmxTarget
from tilefoundry.target.cuda.target import CudaTarget

for declared in (CudaTarget("nvidia.h200_sxm"), AmxTarget()):
    ty = TensorType(shape=(8,), dtype=DType.f32, layout=None, storage="gmem")
    x = Var(type=ty, name="x")
    fn = HirFunction.build(name="k", params=(x,), body=x, return_type=ty)
    module = Module(name="m", functions=(fn,), entry="k", target=declared)
    lowered = HirToTirPass().run(module)
    pf = lowered.functions[0]
    print(f"declared on Module : {module.resolve_target().name:5s} "
          f"{module.resolve_target().architecture.name}")
    print(f"landed on TIR      : {pf.target.name:5s} {pf.target.architecture.name}")
    print(f"  propagated?      : {pf.target is declared}")
    print(f"  codegen groups as: {list(group_functions_by_target(lowered))}\n")
```

### Actual

```
declared on Module : cuda  sm_90
landed on TIR      : cuda  sm_90
  propagated?      : False            # a fresh default, not the declared value
  codegen groups as: ['cuda']

declared on Module : amx   apple_amx
landed on TIR      : cuda  sm_90      # backend silently swapped
  propagated?      : False
  codegen groups as: ['cuda']         # routed to the CUDA emitter
```

### Expected

The declared Target reaches the lowered `PrimFunction`, and an `amx` Module
groups as `['amx']`.

## Notes on detecting this

Both CUDA lines print `sm_90` because the shipped hardware corpus has only
`nvidia.sm90`; the `amx` case is what makes the loss visible. For the CUDA case
the loss cannot be observed with `==` at all — `default_target()` returns a
**fresh equal value** on each call, so only object identity distinguishes
"propagated" from "defaulted".

That is also why this has no runtime symptom: codegen builds for the fallback
arch and the driver PTX-JITs the result, so the kernel still computes correct
answers. There is currently no test asserting target propagation, which is what
allowed it to go unnoticed.

## Root cause

Two sites in `passes/transforms/hir_to_tir.py` build `PrimFunction` with no
`target=`:

- `_lower_function` — one per static body and per mangled dispatch variant
- `_build_dispatch_entry` — the dispatch entry

Since `45bd840` ("give execution context and hardware facts one owner each",
#32) the Target is owned by `Module` (with owner-chain inheritance via
`Module.resolve_target()`), and `hir_to_tir` never consults it.

## Suggested fix

Resolve once in `HirToTirPass.run` via `module.resolve_target()` and thread it
into both sites. A Module declaring no target anywhere should leave the kwarg
off rather than pass `None`, so `PrimFunction`'s own default stays
single-sourced and `group_functions_by_target`'s no-resolved-Target check is not
tripped. Resolving one target per Module also satisfies that function's existing
requirement that a Module's CUDA functions carry identical Target facts.

PR follows.

bigSheep123:
# fix(passes): carry the Module's Target into the lowered TIR

Fixes #63.

## What was wrong

`HirToTirPass` built every `PrimFunction` without `target=`, so the Target
declared by the owning `Module` was discarded and
`field(default_factory=default_target)` — cuda / sm_90 — took over silently.

Because `group_functions_by_target` keys on `fn.target.name`, the failure is not
just a wrong SM arch but a wrong backend: an `amx` Module was routed to the CUDA
emitter.

```
declared on Module : amx   apple_amx
landed on TIR      : cuda  sm_90        # before
  codegen groups as: ['cuda']

declared on Module : amx   apple_amx
landed on TIR      : amx   apple_amx    # after
  codegen groups as: ['amx']
```

## The change

Resolve the target once in `run()` with `Module.resolve_target()` — it walks the
owner chain, so a child Module selected out of a tree still gets the root's
target — and thread it into both construction sites:

- `_lower_function`, for static bodies and mangled dispatch variants
- `_build_dispatch_entry`, for the dispatch entry, so an entry and the variants
  it dispatches to cannot end up on different backends

A Module that declares no target anywhere leaves the kwarg off rather than
passing `None`. Two reasons: it keeps the choice of fallback single-sourced in
`PrimFunction`, and storing `None` would trip `group_functions_by_target`'s own
"no resolved Target" check. Resolving one target per Module also satisfies that
function's existing rule that a Module's CUDA functions all carry identical
Target facts.

`try/except ValueError` around `resolve_target()` matches the idiom
`ir/core/module.py::_reentered` already uses for an undeclared target.

## Tests

Three tests in `tests/passes/test_hir_to_tir.py`, which had no coverage of target
propagation before — the reason this could regress unnoticed:

1. the declared Target reaches a static body, every mangled dispatch variant,
   and the dispatch entry;
2. a child Module lowers against the target it inherits from its root;
3. a Module declaring no target still lowers, on `PrimFunction`'s default.

They assert **object identity**, not equality: `default_target()` returns a fresh
equal value per call, so `==` cannot distinguish "propagated" from "defaulted".
Each test was confirmed to fail against the unfixed pass before being kept.

This is also why no runtime test can catch the bug — the fallback arch still
PTX-JITs to a kernel that returns correct answers.

## Verification

```
pytest tests/ -q     644 passed (main @ e55e4b0)  ->  647 passed (+3 new)
```

Every `.pre-commit-config.yaml` hook clean: ruff, spec-rules, spec-refs,
spec-entropy, forward-references, comment-hygiene, no-machine-paths,
english-only, and clang-format over all 29 C++ files.

### DIFF
--- a/src/tilefoundry/passes/transforms/hir_to_tir.py
+++ b/src/tilefoundry/passes/transforms/hir_to_tir.py
@@ -90,6 +90,7 @@ from tilefoundry.ir.types.shard.shard_layout import (
 from tilefoundry.ir.types.storage import StorageKind
 from tilefoundry.ir.visitor import ExprVisitor
 from tilefoundry.passes.pass_base import ModulePass
+from tilefoundry.target import Target
 from tilefoundry.visitor_registry.registries import (
     hir_lowering_registry,
     register_hir_lowering,
@@ -1400,6 +1401,18 @@ def _lower_single_output(
         lo._items.append(_eval_call(Copy(), (src, out_var)))
 
 
+def _target_kwarg(target: Target | None) -> dict[str, Target]:
+    """``{"target": target}``, or empty when the Module declared none.
+
+    The target picks the backend and, for cuda, the SM arch codegen builds for,
+    so dropping it silently compiles for the fallback arch. Omitting the kwarg
+    leaves ``PrimFunction``'s own default in place; passing ``None`` would store
+    ``None`` and trip ``group_functions_by_target``. The rule:
+    tilefoundry spec core-ir target-inheritance.
+    """
+    return {} if target is None else {"target": target}
+
+
 def _lower_function(
     fn: HirFunction,
     *,
@@ -1411,6 +1424,7 @@ def _lower_function(
     override_name: str | None = None,
     dispatch_groups: "dict[str, tuple[HirFunction, ...]] | None" = None,
     mangled_registry: "dict[str, PrimFunction] | None" = None,
+    target: Target | None = None,
 ) -> PrimFunction:
     """Materialise the HIR `Function(params) -> tensor` as an
     explicit-output-param ``PrimFunction``. The function-end
@@ -1535,12 +1549,15 @@ def _lower_function(
         params=final_params,
         body=body,
         output_count=len(out_vars),
+        **_target_kwarg(target),
     )
 
 
 def _build_dispatch_entry(
     group: tuple[HirFunction, ...],
     mangled_pfs: list[PrimFunction],
+    *,
+    target: Target | None = None,
 ) -> PrimFunction:
     """Build the unmangled entry PrimFunction holding the DispatchCall.
 
@@ -1633,6 +1650,9 @@ def _build_dispatch_entry(
         params=(*entry_params, *out_vars, shape_param),
         body=body,
         output_count=len(out_vars),
+        # Same target as the variants this entry dispatches to, so the entry and
+        # its callees stay on one backend / arch.
+        **_target_kwarg(target),
     )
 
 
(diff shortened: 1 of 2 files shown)


bigSheep123:
Closing to resubmit with corrected commit authorship.

PR #76, opened 2026-08-07
feat(target): add a B200 SXM target on the SM100 architecture

### LINKED ISSUE #77: Only H200 on SM90 can be described, and a second CUDA product has no way to state what it has
# Only H200 on SM90 can be described, and a second CUDA product has no way to state what it has

**Affects:** `main` @ `73c0138`

## Summary

`CudaTarget` composes any architecture and device pair, and external target
support made the services follow the value rather than a name. But the CUDA
package still installs exactly one pair, and the shape of a CUDA architecture and
of a CUDA device is written on the product class: `SM90` declares 11 fields and
two methods, `H200SXM` declares 6 and two more, and every one of them is a
property of *any* CUDA architecture or device. A second product copies both
classes before it can state a single number of its own.

Two facts have nowhere to go even after that copy:

- `target/cuda/facts.py` already projects a `tmem` memory level, and it can only
  ever report `None`, because no architecture value carries a tensor-memory
  capacity. An architecture whose MMA accumulates in tensor memory cannot say so.
- A device can only list the compute DTypes it *has* a peak rate for. There is no
  way to record that a product has no FP4 mode, and no `throughput.f4e2m1` leaf
  in the device schema at all, so a Blackwell part has nowhere to put its
  published FP4 rate and a Hopper part cannot state that it has none.

## Demonstration

```python
import dataclasses

from tilefoundry.analysis.facts import MemoryHierarchyFacts
from tilefoundry.ir.types import DType
from tilefoundry.target import CudaTarget
from tilefoundry.target.cuda import H200SXM, SM90
from tilefoundry.target.hardware import HARDWARE_SPECS

print("installed ids      :", HARDWARE_SPECS.installed_ids())
h = CudaTarget("nvidia.h200_sxm")
print("explicit levels    :", [
    (level.name, level.capacity_bytes)
    for level in h.get_facts(MemoryHierarchyFacts).explicit_levels
])
try:
    CudaTarget("nvidia.b200_sxm")
except Exception as exc:
    print("b200 attempt       :", type(exc).__name__, exc)
print("SM90 own fields    :", len(dataclasses.fields(SM90)))
print("H200SXM own fields :", len(dataclasses.fields(H200SXM)))
try:
    h.device.peak_for(DType.f4e2m1)
except Exception as exc:
    print("h200 fp4 peak      :", type(exc).__name__, exc)
```

### Actual

```
installed ids      : ('apple.amx', 'apple.m2_pro', 'nvidia.h200_sxm', 'nvidia.sm90')
explicit levels    : [('gmem', 141000000000), ('smem', 232448), ('rmem', 262144), ('tmem', None)]
b200 attempt       : UnknownDocumentError no installed hardware document 'nvidia.b200_sxm'; installed: ['apple.amx', 'apple.m2_pro', 'nvidia.h200_sxm', 'nvidia.sm90']
SM90 own fields    : 11
H200SXM own fields : 6
h200 fp4 peak      : ValueError h200_sxm: no dense compute-throughput entry for dtype 'f4e2m1'
```

The `tmem` line is the interesting one: the level is projected with a hardcoded
`None` and a comment saying tensor memory arrives with a later architecture. The
last line is right for H200 for the wrong reason — the rate is absent because
nothing models the leaf, not because the document says the product has no FP4
mode.

## Expected

- A further CUDA product is its two documents plus two value identities. No
  field is redeclared, and no `Target` subclass is needed, since the services
  already follow the value.
- A CUDA architecture states its tensor-memory capacity, or records that it has
  none; the memory-hierarchy projection reports whichever the document says
  instead of a constant.
- A CUDA device states, per compute DType, either a dense peak or that the
  product has no mode for it, so `peak_for` refusing a DType is a statement about
  the hardware rather than a gap in the schema.

## Notes

This blocks describing any Blackwell part, which is what an sm_100 host has to
compile against. I have a change ready for it and will link it here.

bigSheep123:
Closes #77.

## Why

- The CUDA target describes exactly one product, so external target support has no
  second NVIDIA part to be exercised on.
- A B200 has two facts no CUDA value could hold: SM100 accumulates its MMA in a
  tensor-memory store that SM90 does not have, and its tensor cores have an FP4
  mode with a published dense peak.

## What

- `CudaArchitecture` and `CudaDevice` are the whole code side: one value type per
  kind, one schema per kind building it. A CUDA product is its architecture
  document and its device document and nothing else — no `Target` subclass, and no
  architecture or device type of its own, because such a type would carry no
  number and only restate an identity the value already holds.
- An architecture states its tensor-memory capacity or records that it has none,
  which the CUDA memory-hierarchy projection reports as the `tmem` level's
  capacity. A device states a dense peak per compute DType or records the
  absence, so `peak_for` raises for a mode the product does not have.
- `nvidia.sm100` and `nvidia.b200_sxm` install alongside the Hopper pair, so
  `CudaTarget("nvidia.b200_sxm")` composes them, reaches codegen as `sm_100`, and
  projects 148 SMs, its own HBM bandwidth, a 256 KiB tensor-memory level, and a
  dense FP4 peak the H200 has no entry for. Adding a further CUDA product is
  adding two TOML files.
- A provider that needs a fact this shape does not model still subclasses the
  value type and states it from a document of its own; the printer workflow covers
  that path.

Hardware evidence follows the existing rules. The Blackwell tuning guide covers
the per-SM limits; the tensor-core peaks are the published HGX B200 rates divided
by eight and, where the figure is sparse, by two, with the division stated as the
leaf's `conditions`. NVIDIA publishes no SM count and no L2 capacity for this
part, so those are recorded as measured through the CUDA runtime on a B200 SXM
rather than borrowed from a related die.

## Contract

- `docs/spec/target.md` follows the same narrative as the code: `CudaArchitecture`
  is §4.1 and `CudaDevice` is §4.2 under `CudaTarget`, and the four products are
  the documents recorded under them (§4.1.1 SM90, §4.1.2 SM100, §4.2.1 H200SXM,
  §4.2.2 B200SXM). The old §2 and §3 are gone; §4 also states that a further CUDA
  product is two documents rather than a subclass, and §10.2 that a document's
  recorded identity is what it is, not a key into a class table.
- Removed from the public surface: `SM90`, `SM100`, `H200SXM`, `B200SXM`. A caller
  holding one of those types now holds `CudaArchitecture` or `CudaDevice`;
  `CudaTarget("nvidia.h200_sxm")` and every projection are unchanged.
- Renamed: `build_sm90` / `build_h200_sxm` → `build_cuda_architecture` /
  `build_cuda_device`, since one builder now serves every value of its kind.
  Removed: `installed_architecture` / `installed_device`, which named one product
  each once a second was installed; both call sites resolve by ID.
- Both CUDA schemas are `/v2`. They require two leaves a `/v1` document does not
  carry — `memory.tensor.per_cta` and `throughput.f4e2m1` — and
  `SchemaReader.optional_integer` requires the key even when the value is
  recorded unavailable, so an old document would stop loading with a
  missing-fact error rather than a signal that the contract moved.
- Every CUDA architecture value states `tensor_memory_per_cta_bytes`; the external
  `Volta70` provider example gains it as `None`.
- `nvidia.sm100` was standing in as an unknown-document example in the
  resolution-diagnostics workflow; that case now uses an ID that stays unknown.
- One reference in an edited constraint rendered as a link to the Target section
  followed by a literal `0.2`; it now points at the registry subsection it meant.

## Evidence

Run on an eight-GPU B200 SXM host, driver 590.48.01, CUDA 13.1, nvcc 13.1. Every
figure below is from the current tip.

- `pre-commit run --all-files` — all nine hooks pass, no file rewritten.
- `pytest tests/` — 680 passed (on the branch with `main` merged in).
- **The whole real-hardware suite against the new target.** Pointing
  `tests/integration` at `nvidia.b200_sxm` instead of `nvidia.h200_sxm` gives 42
  passed: MMA runtime, handwritten TIR MMA, dynamic CTA launch, dynamic shape
  dispatch, rmsnorm + FP8 quant, and the host-launch group including its
  external-target and negative cases.
- **It is real `sm_100` SASS, not PTX compatibility.** `cuobjdump -lelf` on the
  linked module reports `libmain.1.sm_100.cubin` and `cuobjdump -sass` reports
  `arch = sm_100`.
- Wheel: `python -m pip wheel .` ships `nvidia_sm100.toml` and
  `nvidia_b200_sxm.toml`; installed into a clean environment,
  `CudaTarget("nvidia.b200_sxm")` answers `sm_100 / 148 SMs / tmem 262144`.
- `pytest tests/installed -o python_files='smoke_*.py'` against that wheel — 159
  passed, including the external-V100-provider smoke that reads the architecture
  field this change adds.

## Risk

- No test in the suite requires B200 hardware: the documents and projections are
  checked as values, so a CI host without a Blackwell device is unaffected. The
  runs above are the hardware evidence.
- `arch` is `sm_100`, not `sm_100a`. Nothing emitted today needs the
  architecture-specific variant, but a kernel using the tensor-memory MMA family
  would; selecting it is a codegen concern and not part of this change.
- The B200 SM count and L2 capacity are what the CUDA runtime reports, because
  NVIDIA publishes neither for this part; each leaf names the host and the query
  it came from. The HBM capacity is the published per-GPU figure rather than the
  larger number the part under test reports, so a plan is never sized against
  memory a differently provisioned B200 does not have.

### DIFF
--- a/docs/spec/target.md
+++ b/docs/spec/target.md
@@ -181,6 +181,12 @@ class CudaTarget(Target):
     supplied directly carries no document, so it has no ID or digest and is
     exempt from that check: it is a distinct hardware value rather than a
     revision of an installed one.
+  - `CudaTarget` MUST compose any installed architecture and device pair that
+    declares compatibility. A further CUDA product is its two documents and the
+    two value identities they build
+    ([§10.2](#102-registry-and-resolution)), never a Target subclass:
+    the services and Facts are selected by the value already, so a subclass would
+    have nothing of its own to state.
   - CUDA MUST select the pipeline Scheduler at `thread` and the partition
     Scheduler at `cta` through `get_scheduler`. A CUDA subclass MUST inherit
     those services through ordinary Python inheritance unless it overrides or
@@ -487,6 +493,11 @@ conditions = "No validated number."
     search path, no overlay, and no partial document.
   - A target package MUST register its typed schemas and installed documents as
     an import side effect, into the same shared registry.
+  - A schema MAY build more than one value class when every document it validates
+    states the same fact paths. The identity a document declares MUST select the
+    class, and an identity the schema builds no class for MUST raise naming the
+    identities it does build. That selection reads the document's own recorded
+    content, so it is not a dispatch on a Target name.
   - A typed schema MUST validate exact fact paths, value types, units, required
     fields, and cross-field invariants, and MUST reject any leaf the document
     carries that the schema does not model, so a misspelled key cannot become
@@ -636,3 +647,40 @@ class CudaDevice(Device):
     ([§10](#10-installed-hardware-resources)). Selecting a different installed
     document by ID is not an override; supplying a partial or edited number
     without a document behind it is, and is not admitted.
+
+## 14. `SM100`
+
+```python
+class SM100(CudaArchitecture):
+    """SM100 compilation identity and structural capabilities."""
+```
+
+- constraints:
+  - `name` MUST be `sm_100`, the architecture identity CUDA compilation uses.
+  - SM100 MUST carry the CUDA architecture facts and add no field of its own
+    ([§12](#12-cudaarchitecture)).
+  - SM100 MUST report `f4e2m1` as a compute DType, because its MMA takes 4-bit
+    operands directly. `f8e8m0` MUST NOT be reported as one: it scales those
+    operands rather than being multiplied.
+  - SM100 MUST state a tensor-memory capacity, and its instruction capabilities
+    MUST name the MMA family that accumulates there rather than the SM90 family,
+    which this architecture does not run.
+  - Device-frequency-dependent FLOP/s values MUST NOT be stored on SM100.
+
+## 15. `B200SXM`
+
+```python
+class B200SXM(CudaDevice):
+    """One B200 SXM device."""
+```
+
+- constraints:
+  - B200SXM MUST carry the CUDA device facts and add no field of its own
+    ([§13](#13-cudadevice)).
+  - `peak_for` MUST expose a dense integer FLOP/s entry for each of `f32`,
+    `f16`, `bf16`, `fp8e4m3`, and `f4e2m1`, each value taken from the installed
+    document.
+  - Resource facts the vendor does not publish for this product, its SM count
+    and its L2 capacity among them, MUST be recorded as measured on the
+    described host rather than estimated or borrowed from a related part
+    ([§10.1](#101-document-envelope)).
--- a/docs/spec/target.md
+++ b/docs/spec/target.md
@@ -253,6 +253,9 @@ class CudaArchitecture(Architecture):
     capacity per SM and per CTA, and register-file capacity per SM. These are
     properties of the microarchitecture, so every product built on it shares
     them, and a device MUST NOT restate them.
+  - It MUST NOT carry a compute-throughput rate. A FLOP/s figure depends on the
+    clock of one product, so it is a device fact ([§4.2](#42-cudadevice)) even
+    though the instruction it rates is the architecture's.
   - `unified_l1_shared_per_sm_bytes` MUST be the size of the one physical block
     the shared-memory carveout and the L1 data cache are both taken from, and
     MUST be at least `shared_memory_per_sm_bytes`. The architecture MUST NOT
@@ -288,7 +291,6 @@ The installed `nvidia.sm90` architecture document.
     compute DTypes.
   - It MUST record no tensor-memory capacity: the SM90 MMA accumulates in the
     register file, so there is no separate store to size.
-  - Device-frequency-dependent FLOP/s values MUST NOT appear in it.
 
 #### 4.1.2 SM100
 
@@ -302,7 +304,6 @@ The installed `nvidia.sm100` architecture document.
   - It MUST record a tensor-memory capacity, and its instruction capabilities
     MUST name the MMA family that accumulates there rather than the SM90 family,
     which this architecture does not run.
-  - Device-frequency-dependent FLOP/s values MUST NOT appear in it.
 
 ### 4.2 `CudaDevice`
 
@@ -573,7 +574,7 @@ compatibility, never a single combined record.
 
 ```toml
 [spec]
-schema = "tilefoundry.cuda.device/v1"
+schema = "tilefoundry.cuda.device/v2"
 kind = "device"
 id = "nvidia.h200_sxm"
 
@@ -622,13 +623,16 @@ conditions = "No validated number."
 - constraints:
   - `HardwareSpecRegistry` MUST resolve documents by exact ID. There MUST be no
     search path, no overlay, and no partial document.
+  - A schema name carries a version. Requiring a leaf a previous version did not
+    MUST take a new version, because a document written against the old one no
+    longer loads and the failure is otherwise a missing fact rather than a
+    contract that moved.
   - A target package MUST register its typed schemas and installed documents as
     an import side effect, into the same shared registry.
-  - A schema MAY build more than one value class when every document it validates
-    states the same fact paths. The identity a document declares MUST select the
-    class, and an identity the schema builds no class for MUST raise naming the
-    identities it does build. That selection reads the document's own recorded
-    content, so it is not a dispatch on a Target name.
+  - A schema MAY validate the documents of several products when they state the
+    same fact paths, and MUST build one value type from all of them. A product is
+    what its document records, so a schema MUST NOT select a type by the identity
+    a document declares.
   - A typed schema MUST validate exact fact paths, value types, units, required
     fields, and cross-field invariants, and MUST reject any leaf the document
     carries that the schema does not model, so a misspelled key cannot become
--- a/docs/spec/target.md
+++ b/docs/spec/target.md
@@ -101,78 +101,35 @@ class Device:
 ## 2. `SM90`
 
 ```python
-class SM90:
+class SM90(CudaArchitecture):
     """SM90 compilation identity and structural capabilities."""
-
-    name: str
-    supported_compute_dtypes: tuple[DType, ...]
-    instruction_capabilities: tuple[str, ...]
-    max_threads_per_cta: int
-    max_threads_per_warp: int
-    max_warps_per_cta: int
-    max_resident_ctas_per_sm: int
-    shared_memory_per_sm_bytes: int
-    shared_memory_per_cta_bytes: int
-    unified_l1_shared_per_sm_bytes: int
-    registers_per_sm_32bit: int
-
-    def supports_compute_dtype(self, dtype: DType) -> bool: ...
-
-    def topology_limit(self, name: str) -> int: ...
 ```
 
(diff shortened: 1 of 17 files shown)


docs/spec/target.md
```diff
@@ -101,78 +101,35 @@ class Device:
 ## 2. `SM90`
```
zhen8838: This spec can be modified accordingly, and the new narrative should be
```md
###1.3 CudaTarget
####1.3.1 CudaArchitecture
1. SM90
2. SM100
####1.3.2 CudaDevice
1. H200SXM
2. B200SXM
```

src/tilefoundry/target/cuda/device.py
```diff
@@ -47,4 +51,14 @@ def peak_for(self, dtype: DType) -> int:
             ) from None
 
 
-__all__ = ["H200SXM"]
+@dataclass(frozen=True)
+class H200SXM(CudaDevice):
```
zhen8838: There is something wrong with the design of CUDA device and Architecture. First of all, they only have one packaging class in the code, and there is no specific implementation. Their specific parameters are written in Toml. Should to redesign it.

zhen8838 (changes requested):

bigSheep123:
Both points taken, pushed as `refactor(target): make a CUDA product its two documents alone`.

**The device and architecture design.** You are right, and the answer is that
those classes should not exist. `SM90`, `SM100`, `H200SXM`, and `B200SXM` carried
no number of their own — every one comes from the installed TOML — so each class
only restated an identity its value already held, and the schema needed a table
mapping a recorded identity back to a class to build it. All of that is removed.
`CudaArchitecture` and `CudaDevice` are now the whole code side: one value type
per kind, one schema per kind building it, and a product that differs from
another only in what its documents record. Adding a CUDA product is adding two
TOML files, and the net change to the package is smaller than before this commit.

The extension point that mattered is kept: a provider that needs a fact this
shape does not model subclasses the value type and states it from a document of
its own, which the python-printer workflow exercises.

**The spec narrative.** Restructured to your outline — `CudaArchitecture` and
`CudaDevice` under `CudaTarget`, each product recorded beneath the type it builds:

```
## 4. CudaTarget
### 4.1 CudaArchitecture
#### 4.1.1 SM90
#### 4.1.2 SM100
### 4.2 CudaDevice
#### 4.2.1 H200SXM
#### 4.2.2 B200SXM
```

Two notes on how I numbered it. I kept `CudaTarget` at §4 rather than moving it
to §1.3, because `core-ir.md` and `parser.md` both link `target.md#4-cudatarget`
and the spec-refs rule is that a number is an address; the old §2 and §3 are now
gaps, which `SPEC-RULES.md` admits. And the products are recorded as documents
rather than as types, since that is what they are after this change. If you want
the three concrete targets to sit symmetrically under §1 — `1.3 CudaTarget`,
`1.4 CpuTarget`, `1.5 AmxTarget` — say so and I will do it as a follow-up commit
here; it touches the AMX and CPU sections and the links in the two other specs,
which is why it is not folded into this one.

**Re-validated after the redesign**, on an eight-GPU B200 SXM host (driver
590.48.01, CUDA 13.1):

- `pre-commit run --all-files` — nine hooks pass, nothing rewritten
- `pytest tests/` — 657 passed
- `tests/integration` pointed at `nvidia.b200_sxm` — 42 passed, and
  `cuobjdump` on the linked module reports `libmain.1.sm_100.cubin` /
  `arch = sm_100`
- wheel built, installed into a clean environment, `pytest tests/installed` — 159
  passed

bigSheep123:
Thanks for merging `main` in. I then went over the whole change again and fixed
nine things; two of them were real defects rather than tidying.

**The B200 device document described two products at once.** Its HBM capacity was
a reading taken off one 192 GB part, while every rate in the same document is
derived from the published HGX B200 figures — the 180 GB part. That value reaches
`PartitionFacts.memory_capacity_bytes` and the `gmem` level, so a plan for the
part those rates describe would have been sized against 11.5 GB of HBM it does not
have. The capacity now comes from the same published source as the rates
(1,440 GB / 8), and the conditions say why the larger reading is not used.

**Both CUDA schemas require a leaf they did not before** — `memory.tensor.per_cta`
and `throughput.f4e2m1` — and `SchemaReader.optional_integer` requires the key
even when the value is recorded unavailable. So a `/v1` document authored against
the old contract stops loading, with a missing-fact error rather than a signal
that the contract moved. Both schemas are `/v2` now, and §10.2 states the rule.

The rest:

- `nvidia.sm100` declared `devices = ["nvidia.b200_sxm"]`. A third SM100 product
  would then have required editing the architecture document — changing its digest,
  and so the recorded provenance of every artifact already built against it —
  which contradicts §4's "a product is its two documents". The device side declares
  the pair, as `nvidia.sm90` already does.
- §10.2 still described a schema selecting a value class by a document's identity.
  That is the mechanism the previous commit removed, and the opposite of what §4.1
  says; it is rewritten to state that a schema builds one value type for every
  document it validates.
- `installed_sm90` / `installed_h200_sxm` named one product each — the naming this
  change removes everywhere else — and left the new product with no accessor. Both
  are gone; the two call sites resolve by ID.
- The FP4 leaf's derivation cited "72 PFLOP/s dense FP4"; the row also prints
  144 PFLOP/s sparse, so it now names both and there is no way to read the divided
  figure as the FP8 row's sparse rate.
- "Device-frequency-dependent FLOP/s values MUST NOT appear" was stated under two
  products when it holds for every architecture; it is one constraint on the type.
- Two entries in the no-Python-defaults loop re-asserted the same two value types,
  and two docstrings carried acceptance-criteria tags pointing at a plan this
  repository does not carry. Both removed. `__all__` is sorted again.

Four things I looked at and deliberately did not change, in case you disagree:

- **`memory_hierarchy` reads `tensor_memory_per_cta_bytes` unconditionally**, so an
  out-of-tree `Architecture` that is not a `CudaArchitecture` raises `AttributeError`
  on upgrade. Guarding with `getattr(..., None)` would report a level without a
  capacity for hardware nobody described, which §11 forbids ("MUST NOT fall back to
  ... built-in hardware values"). §4.1 now states the requirement instead, and the
  in-tree provider example shows the one-line change.
- **No builder check that a tensor-memory architecture names the matching MMA
  family.** Enforcing it means testing capability strings by name in the builder,
  which is the dispatch this PR removes; it stays a documented rule like the
  `origin` honesty rules in §10.1.
- **`_roofline_duration_ns` calls `peak_for` unguarded**, so registering an FP4 atom
  would raise on a device with no FP4 mode instead of listing no candidate. Real,
  but it is in `atoms.py`, untouched here, and unreachable until such an atom exists.
- **§2 and §3 are now numbering gaps.** Repairing them means renumbering, which the
  spec rules forbid because other documents link these anchors.

Re-validated at `fcb2610`: `pre-commit run --all-files` clean, `pytest tests/`
680 passed, `tests/integration` pointed at `nvidia.b200_sxm` 42 passed with
`cuobjdump` reporting `libmain.1.sm_100.cubin` / `arch = sm_100`, and the wheel
installed into a clean environment with `pytest tests/installed` 159 passed.

zhen8838 (approved):

PR #78, opened 2026-08-08
feat(shard): collapse a split axis locally and state mesh positions once

zhen8838:
## Why
- A tiled program could not be analysed. `local_type_of` rejected a `ceildiv(S, T)` tile-count axis — the shape every real kernel works in — so every cost handler died on it even though the projected shape was fully static.
- A mesh stated its position count twice, once in `topologies` and once in `layout`, and seven readers each picked a side.
- An interrupted test compile left a `torch.utils.cpp_extension` lock behind. `FileBaton.wait()` has no timeout, so every later run hung with no diagnostic and no failure.

## What
- A `Split`-bound axis projects to a local extent of 1 rather than dividing, which is what `docs/spec/shard.md` §7.1.1 already states. The concreteness check moves after the fold, so a dynamic tile count yields a static local shape and a cost.
- `Mesh` becomes a plain frozen record of `(topologies, layout, names)` with no `__post_init__`; consistency is a named predicate where it is needed. Position counts are read from the layout. Parser, printer, analysis, specs and tests migrate with it.
- Test CUDA extensions build inside the worktree, and the pytest controller clears abandoned locks before collection.

## Contract
- `Mesh(topology=...)` is removed and `topologies` is always a tuple: `with Mesh(("cta",), layout=(128,))`.
- `MeshAxis`, `Mesh.topology`, `Mesh.topology_domain()` and attribute-style axis access are removed. In layout sugar `m.axis` stays parser syntax, resolved to a `(mesh, axis index)` pair.
- A mesh spanning more than one topology is refused when a per-level position count is asked of it, rather than answered with a guess.
- A call whose types state no placement now reports `grid_units=1` instead of the function's declared topology size.

## Risk
- The placement change alters reported attribution only. The `qwen3_1_7b` decoder layer's makespan is unchanged — 21098 ns at `ctx_len=1`, 29009 ns at `ctx_len=1024` — because the old 132 units equalled the H200 CTA capacity, so neither the old nor the new extent needs an extra wave and durations did not move.
- `schedule/partition` keeps its own extent readers; that level is withdrawn by later work and is left alone here.


tests/ir/types/test_mesh.py
```diff
@@ -0,0 +1,116 @@
...
+from tilefoundry.ir.types.shard.scope_match import (
+    mesh_scope_matches_required_scope,
+    states_consistent_positions,
+)
+from tilefoundry.ir.types.shard.shard_layout import ShardLayout
+from tilefoundry.parser.sugar import parse_shard_layout_sugar
+from tilefoundry.schedule.partition.problem import _placement_relation
+
+
+def _mesh(name: str, topology_size: int | None, layout_shape: tuple[int, ...]) -> Mesh:
```
zhen8838: `make_mesh` should be used for testing.

tests/test_conftest.py
```diff
@@ -0,0 +1,37 @@
+from __future__ import annotations
+
+import os
+from pathlib import Path
+from types import SimpleNamespace
+
+
+def test_pytest_configure_isolates_extensions_and_leaves_worker_locks(
```
zhen8838: remove this test

PR #111, opened 2026-08-23
refactor(parser): migrate to unified function API

zhen8838:
## Why
- Replace the legacy parser with the unified Pattern/Rule parser without compatibility readers, retarget metadata, or hidden dtype promotion.

## What
- Defer in-module function parsing until child Modules are available, then resolve declarations in source order.
- Keep explicit scalar casts on the ordinary `CallPattern -> ConstantPattern -> Cast Call` path.
- Treat UMAT-only Call results as scalar expressions rather than GMEM placement values.
- Move retained legacy parser coverage under `tests/parser` and update the parser spec.

## Contract
- `parse_function(fn, context)` remains the only parser entry point.
- Binary dtype matching stays strict; authored mismatches use explicit casts.
- Parser Rules do not inspect concrete `Cast(Constant)` combinations.

## Risk
- Source full verification: 887 passed, 4 skipped.
- Installed full verification: 175 passed, 0 skipped.
- JUnit, coverage, commands, and outputs are retained under `test_results/parser-scalar-call-*-full-r1/`.


src/tilefoundry/ir/core/module.py
```diff
@@ -23,6 +24,24 @@
...
+
+
+def _authored_child_call(call):
+    record = get_metadata(call, _ModuleCallee)
+    return None if record is None else record.owner
+
+
+from tilefoundry.ir.hir._call_binding import set_authoring_reader  # noqa: E402
+
+set_authoring_reader(_authored_child_call)
```
zhen8838: Why is this code? In this way, it will be executed when it becomes import.

src/tilefoundry/parser/ast_pattern.py
```diff
@@ -0,0 +1,1529 @@
...
+
+from tilefoundry.ir.core import VerifyError
+
+T = TypeVar("T")
+_RETURN_TYPE = "<return_type>"
+_TYPE_INFER_CONTEXT = "<type_infer_context>"
+
+
+@lru_cache(maxsize=1)
+def _runtime() -> SimpleNamespace:
```
zhen8838: Isn’t this paragraph the compatible code used for development? Why don‘t you delete it?

tests/ir/types/test_mesh.py
```diff
@@ -20,7 +20,7 @@
     states_consistent_positions,
 )
 from tilefoundry.ir.types.shard.shard_layout import ShardLayout
-from tilefoundry.parser.sugar import parse_sugar
+from tilefoundry.ir.types.shard.sugar import parse_sugar
```
zhen8838: Why is Parser Sugar imported from there? Shouldn’t it be a part of Parser? Then I don‘t think the position of this test is reasonable.

zhen8838:
CI run 32637567894 completed. Pre-commit passed; the test job reproduced the same five baseline failures as the clean-main verification (879 passed, 5 failed, 4 skipped), with no new M11-specific failures. The run stopped before wheel/install checks because the repository CI treats baseline failures as job failure.

Local fresh evidence for commit `86de9ce`: `test_results/parser-migration-m11-20260823e/` (JUnit `junit.xml`, coverage XML, command/output artifacts). M11-focused blast-radius tests passed: 27 passed.

zhen8838:
Follow-up CI run 32638014364 on docs commit `52e6023` also completed: pre-commit passed and the test job reproduced the same five baseline failures (no new parser/M11 failure). The wheel/install stages were skipped by the workflow after the repository test job failed. PR remains open for human review; no merge was performed.

Issue #122, opened 2026-08-25
fix(analysis): resolve cross-scope predecessors in the performance timeline


zhen8838:
`--performance` crashes with an internal `KeyError` when a value produced inside
an inner loop is consumed one scope up, and the producer is a call whose own
body has a loop.

```
  File ".../tilefoundry/analysis/timeline.py", line 279, in build
    ready = max(
  File ".../tilefoundry/analysis/timeline.py", line 280, in <genexpr>
    (occurrences[key].end_ns for key in occurrence.predecessors),
     ~~~~~~~~~~~^^^^^
KeyError: 131376854249216
```

`--compute-cost --memory --roofline` on the same program are fine; only
`performance` fails, and it fails before any report is written.

### Where it comes from

`occurrences` is the scope-local dict `build()` is filling, but
`occurrence.predecessors` holds ids `representative()` resolved, which can name
an expression that is not in *that* dict:

```python
def representative(producer: int, scope: _Chain) -> int | None:
    host = hosts[producer]
    if host == scope:
        return producer
    if len(host) > len(scope) and host[: len(scope)] == scope:
        return host[len(scope)]          # <-- an id from a deeper chain
    return None
```

Then:

```python
ready = max(
    (occurrences[key].end_ns for key in occurrence.predecessors),
    default=origin_ns,
)
```

`occurrences[key]` assumes every predecessor is a direct occurrence of the
scope being built. A `.get(key)` with the missing case skipped would not crash,
but the interesting question is which chain the id belongs to -- the value here
is two loops deeper than its consumer, so `host[len(scope)]` names a loop that
this scope does not hold directly.

### Reproducer

Complete program, 40 lines. `NB=1` and `GL=2` are enough; the crash needs the
producer to be a *call* whose body has a loop -- the same shape written with a
plain `h = h + h` inside the inner loop is accepted.

```python
from __future__ import annotations

from tilefoundry import func, module
from tilefoundry.dsl import DimVar, Mesh, Tensor, tf
from tilefoundry.dsl.tf import *  # noqa: F401, F403

from tilefoundry.ir.types.shard import Topology
from tilefoundry.target import CudaTarget

GRID, HP, NHS, RTG, NB, GL = 128, 128, 16, 256, 1, 2
S = DimVar("seq_len", 0, 8193)
_H200, _CTA = CudaTarget("nvidia.h200_sxm"), Topology("cta", GRID)


@module(entry="walk", target=_H200, topologies=(_CTA,))
[...]

Issue #121, opened 2026-08-25
fix(parser): name the cause when concat, a docstring, or a bare loop variable is refused


zhen8838:
Three authoring forms are refused with a message that names neither the
construct nor what to write instead. Each cost a debugging round to identify,
and in each case the fix in the source is one character to one line — so the
whole cost was the diagnostic.

`tilefoundry 0.0.2.dev43+g45032b967` (45032b9), target `nvidia.h200_sxm`.

### 1. `tf.concat` with a list

```python
return tf.concat([xs[:, 32:64], xs[:, 0:32]], axis=1)
```

```
expression.attribute: pattern did not match (role 'inputs') at case.py:22:26
```

The parser wants varargs — `tf.concat(a, b, axis=1)`. The list form is how
torch, numpy, jax and tvm all spell it, so it is the first thing an author
tries. Nothing in the message says `concat`, says `list`, or says `varargs`.
Either accept the sequence form or say `tf.concat takes its inputs as varargs,
not a sequence`.

### 2. A docstring in a `@func` body

```python
@func
def f(x: Tensor[(N, W), "f32"]) -> Tensor[(N, W), "f32"]:
    """What this stage does."""
    with Mesh(("cta",), layout=(N,), names=("u",)) as m:
        return tf.reshard(x, (N @ m.u, W), "smem")
```

```
expression.attribute: pattern did not match (role 'statement_value') at case.py:20:5
```

Every other Python function may carry one, and the shipped models put their
prose in `#` comments — which reads like a style choice until you find out it
is enforced. Dropping a leading string constant from the body is a two-line
change in the block pattern; refusing it with `a @func body may not hold a
docstring; use # comments` is a one-line change. Either is better than
pointing at column 5 of a string literal.

### 3. A bare loop variable as a slice bound

```python
for t in tile(S, N):
    xs = tf.reshard(x[:, t:t + N, :], (1, N @ m.u, W), "smem")
```

```
tilefoundry: error: dim_expr: expected int, DimVar, or Expr, got slice
```

`base = t + 0` then `x[:, base:base + N, :]` is accepted and identical.
`docs/tutorial/authoring.md` writes `base = start + 0` in `Stage5_CachePrepared`
without saying why, which is the only hint available. The message describes a
slice arriving where a dim was expected, which is the opposite of the problem
(a loop variable arrived where a dim was expected, inside a slice). Accepting
the bare variable is the right fix; failing that, say `a loop induction
[...]

Issue #120, opened 2026-08-25
fix(dsl): make X @ mesh.axis mean one thing in reshard layouts and Tensor types


zhen8838:
## Why
- `X @ mesh.axis` means two different things in the two places an author writes it. In a `reshard` layout it keeps the tensor's rank; in a `Tensor[...]` type it splits the axis into two. So the two cannot meet:

```
tilefoundry: error: Binary: Binary ADD: cannot broadcast shapes (1, 16, 8192) and (1, 16, 128, 64)
```

- Both operands were written with the same sugar over the same mesh in the same body.

## What
- `tf.reshard(x, (1, 16, 8192 @ m.u), "smem")` has type `Tensor[(1, 16, 8192), ..., ((1, 16, 128 @ cta.u, 64), ...)]` — logical rank 3, the split recorded in the layout.
- `tf.zeros(Tensor[(1, 16, 8192 @ m.u), "f32", "smem"])` has logical shape `(1, 16, 128, 64)` — the split is in the *shape*.
- Consequence: an accumulator seeded with `tf.zeros` cannot be added to, or accumulated into by, anything produced by a resharded op, which is the ordinary way to open a reduction. `docs/tutorial/authoring.md`'s `Stage3_Fused` works only because every operand there is `tf.zeros`/`tf.full_like` of the same seed, so the two forms never meet.

## Contract
- One notation, one meaning. Whichever way it is resolved, `reshard` and `Tensor[...]` should agree on whether `@` changes rank.

## Risk
- If `Tensor[...]` is changed to match `reshard`, programs that index the split axes of a `tf.zeros` result break. If `reshard` is changed to match `Tensor[...]`, most sharded programs break. A third option — reject the sugar in one of the two positions — is the smallest change and at least turns a broadcast error into a statement about the notation.
- Workaround today: open the accumulator with the first iteration's `matmul` instead of `tf.zeros`, which is what we ended up doing everywhere.

## Reproducer

```python
from __future__ import annotations
from tilefoundry import func, module
from tilefoundry.dsl import Mesh, Tensor, tf
from tilefoundry.dsl.tf import *  # noqa: F401, F403
from tilefoundry.ir.types.shard import Topology
from tilefoundry.target import CudaTarget

N = 128
_H200, _CTA = CudaTarget("nvidia.h200_sxm"), Topology("cta", N)


@module(entry="f", target=_H200, topologies=(_CTA,))
class M:
    @func
[...]

Issue #119, opened 2026-08-25
fix(parser): fold a body-local constant expression before the shard divisibility check


zhen8838:
## Why
- A body-local name for a constant expression is not folded. `width = A + B` with `A = B = 4096` stays an IR `Binary` Call, so `8192 @ m.u` is not recognised as divisible by a 128-position mesh and the program is refused.
- The refusal prints an internal repr rather than the expression the author wrote:

```
tilefoundry: error: shard_layout_local_shape: layout dim 2 (Call(type=TensorType(shape=(),
dtype=IntegerDType(name='i64', bit_width=64, signed=True), layout=None,
storage=<StorageKind.UMAT: 6>), target=Binary(kind=<BinaryKind.ADD: 'add'>),
args=(Constant(...value=4096), Constant(...value=4096)))) and mesh axis 0 extent 128
do not have a decidable divisibility relation; bind symbolic dimensions before local projection
```

## What
- The same arithmetic written at module level (`W = A + B` outside the class) is folded and accepted.
- The same arithmetic written inline in the layout tuple (`(1, 16, (A + B) @ m.u)`) is also accepted.
- Only *naming* it inside the body fails, so the constant reaches the shard check unfolded.

## Contract
- Two spellings of one extent should be one extent. Today a body-local name changes whether a legal shard layout is legal.
- The message tells an author to "bind symbolic dimensions" for a dimension that is not symbolic; `4096 + 4096` is fully static.

## Risk
- Two separable fixes: fold static integer `Binary` in the body (removes the refusal), and print the authored expression instead of the node repr (removes the misleading advice). The second is worth doing even if the first is not, because the same message is reachable from a genuinely open dimension.

## Reproducer

```python
from __future__ import annotations
from tilefoundry import func, module
from tilefoundry.dsl import Mesh, Tensor, tf
from tilefoundry.dsl.tf import *  # noqa: F401, F403
from tilefoundry.ir.types.shard import Topology
from tilefoundry.target import CudaTarget

N, A, B = 128, 4096, 4096
_H200, _CTA = CudaTarget("nvidia.h200_sxm"), Topology("cta", N)


@module(entry="f", target=_H200, topologies=(_CTA,))
class M:
    @func
    def f(x: Tensor[(1, 16, 8192), "f32"]) -> Tensor[(1, 16, 8192), "f32"]:
        with Mesh(("cta",), layout=(N,), names=("u",)) as m:
            width = A + B                                  # <-- refused
[...]

PR #123, opened 2026-08-25
fix(ir): correct model analysis and parser surfaces

zhen8838:
## Why
- #117, #119, and #121 exposed analysis and parser-surface defects in authored HIR models.

## What
- Correct MatMul work accounting and add explicit MK/KM and KN/NK operand layouts.
- Normalize integer scalar HIR expressions; fix parser sequences, slices, and generated spec sections.

## Contract
- MatMul defaults remain MK/KN; access relations and cost share one logical-axis mapping.

## Risk
- #118, #120, and #122 remain unresolved; #122 stays in a local plan.


docs/spec/core-ir.md
```diff
@@ -493,10 +494,15 @@ class ParamDef:
     pattern: Pattern | None = None
     optional: bool = False
     default: Any = MISSING
+
+class VariadicList:
```
zhen8838: We can't add this object to core IR. This should be just a temporary object in the parser.

docs/spec/hir.md
```diff
@@ -859,15 +864,15 @@ class Stack(Op):
     Attributes:
         inputs: input; variadic tensors to stack.
         axis: attribute; inserted result axis.
-        is_variadic: attribute; Whether the input parameter consumes all args.
     """
 
-    inputs: Tensor
+    inputs: VariadicList
```
zhen8838: Here, `Tuple[Tensor]` is directly used as a type prompt for variable length input.

zhen8838: At the same time, you need to check whether `Tuple[xxx]` can be used as a general variable length input type prompt for parser?

docs/spec/parser.md
```diff
@@ -104,8 +101,8 @@ loop-carry-statement  ::= expression '=' expression
                           | 'for' name 'in' expression ':' loop-carry
                           | statement
 loop-carry            ::= (loop-carry-statement (newline loop-carry-statement)*)?
-loop-header           ::= 'for' identifier 'in' ('tile' | 'range') '(' expression (',' expression)*
-                          ')' ':' loop-carry
+loop-header           ::= 'for' identifier 'in' identifier '(' (expression | name '=' expression)
```
zhen8838: The iterator name of this loop is not displayed. There should be a special iterator name pattern, and then serialize it as ('tile' | 'range') similar to Python's current literal[str]

src/tilefoundry/analysis/timeline.py
```diff
@@ -223,13 +223,27 @@ def _schedule(
     for expr in values:
         if id(expr) in hosts:
             direct.setdefault(hosts[id(expr)], []).append(expr)
+    held: dict[_Chain, set[int]] = {
```
zhen8838: You said that the modification here was wrong. Why didn't you revert it?

src/tilefoundry/ir/types/shard/shard_layout.py
```diff
@@ -160,31 +160,61 @@ def shard_layout_local_shape(
             k = attr.axis
             if not (0 <= k < len(local)):
                 continue
+            from ..shape_helpers import static_dim_value  # noqa: PLC0415
+
             mesh_ext = mesh_shape[mesh_axis_idx]
+            here = static_dim_value(local[k])
```
zhen8838: Why is this modification? First of all, you can't import in the function.

src/tilefoundry/ir/types/shape_helpers.py
```diff
@@ -29,7 +29,43 @@ def static_dim_value(dim):
         return None
     if isinstance(dim, Constant) and isinstance(dim.value, int) and not isinstance(dim.value, bool):
         return int(dim.value)
-    return None
+    return _folded_dim_value(dim)
```
zhen8838: This is also a wrong practice and needs to be retreated.

src/tilefoundry/parser/grammar_render.py
```diff
@@ -492,12 +496,16 @@ def render(self, root: ElementPattern[Any]) -> str:
...
+    root: ElementPattern[Any] | None = None,
+    *,
+    extra_roots: tuple[ElementPattern[Any], ...] = (),
+) -> str:
     if root is None:
         from .pattern_nodes import FunctionPattern
 
         root = FunctionPattern()
-    return RenderVisitor().render(root)
+    return RenderVisitor().render(root, extra_roots)
```
zhen8838: Why do you need extra root? Wrong design

docs/spec/hir.md
zhen8838: The parser now recognizes Tuple[T] generically via get_origin/get_args; it is not hard-coded to Concat or Stack.

zhen8838: Changed Concat and Stack to use Tuple[Tensor] as the input type prompt in 1385518.

src/tilefoundry/parser/grammar_render.py
zhen8838: Removed extra_roots. Parser spec generation now traverses only FunctionPattern and updates only the two explicitly marked generated sections.

docs/spec/core-ir.md
zhen8838: Removed the Core IR VariadicList marker in 1385518. Variadic surface state now remains parser-local.

src/tilefoundry/ir/types/shard/shard_layout.py
zhen8838: Reverted the function-local import and all ShardLayout-side dimension evaluation.

src/tilefoundry/analysis/timeline.py
zhen8838: Reverted the complete #122 timeline workaround and its weak regression. #122 is excluded from this PR and remains in the local plan.

src/tilefoundry/ir/types/shape_helpers.py
zhen8838: Removed _DIM_FOLD and _folded_dim_value. Integer scalar HIR arithmetic is now handled by the existing ISL normalization visitor and entered through ShapeDimRule.

docs/spec/parser.md
zhen8838: Added a dedicated LoopIteratorPattern whose executable grammar is the tile/range literal choice, then regenerated the spec.

docs/spec/parser.md
```diff
@@ -194,17 +196,21 @@ function              ::= 'def' name '(' signature ')' ('->' return-type)? ':' b
...
+| index_slice | subscript_index | TileWindowSliceBoundRule | A tile window cannot be used as a slice bound. | src/tilefoundry/parser/pattern_nodes.py |
 | layout | tensor_optional_slot | LayoutPositionRule | A layout must be legal for its parser position. | src/tilefoundry/parser/ast_pattern.py |
 | layout | tensor_optional_slot | LayoutShapeRule | A layout must have a valid non-boolean shape. | src/tilefoundry/parser/ast_pattern.py |
 | op_call | expression | CallBindingRule | A call must bind its arguments into a Call tuple. | src/tilefoundry/parser/pattern_nodes.py |
 | op_call | expression | CallExpectedTypeRule | A call's inferred type must satisfy the expected expression type. | src/tilefoundry/parser/pattern_nodes.py |
 | op_call | expression | CallTypeInferenceRule | A call's result type must be inferred from its binding. | src/tilefoundry/parser/pattern_nodes.py |
+| op_call | expression | CallVariadicInputFormRule | A variadic call must use one explicit list, tuple, or supported static list comprehension. | src/tilefoundry/parser/pattern_nodes.py |
```
zhen8838: Why is the same rule printed 3 times?

src/tilefoundry/visitor_registry/op_cost.py
```diff
@@ -115,21 +113,20 @@ def _serviced(call: Call, ctx: CostContext, kind: str) -> Cost:
...
+    care which.
     """
     lhs, rhs = _input_types(call, ctx)
     output = _output_type(call, ctx)
     if not all(isinstance(type, TensorType) for type in (lhs, rhs, output)):
         raise ValueError("MatMul cost requires tensor inputs and output")
-    m, k, n = lhs.shape[-2], lhs.shape[-1], rhs.shape[-1]
-    batch = math.prod(output.shape[:-2])
-    flops = 2 * batch * m * k * n
+    flops = 2 * numel(output) * lhs.shape[-1]
```
zhen8838: Isn't this right? Isn't matmul's FLOPS 2 m k n?

tests/analysis/test_isl_utility.py
```diff
@@ -49,6 +51,18 @@ def test_normalize_dim_uses_isl_affine_normal_form():
     assert normalize_dim(quotient) == expected
 
 
+def test_normalize_dim_accepts_integer_scalar_hir_arithmetic():
```
zhen8838: There is no need to add this test. It was measured by the parser test.

tests/models/qwen3_5_35b_a3b/model.py
```diff
@@ -130,7 +130,7 @@ def conv_step(
         # closes on this token, so the whole convolution is one multiply against
         # the kernel and one reduction over it. Channels do not mix -- that is
         # what depthwise means here, and it is why no matmul appears.
-        window = tf.concat(conv_state, entry, axis=2)
+        window = tf.concat([conv_state, entry], axis=2)
```
zhen8838, line 133: tutorial authoring那边是否也得适配一下？

tests/parser/test_spec.py
```diff
@@ -0,0 +1,53 @@
...
+
+<!-- parser-constraints:start -->
+stale constraints
+<!-- parser-constraints:end -->
+
+kept after constraints
+"""
+
+
+def test_parser_spec_generation_only_replaces_marked_sections() -> None:
```
zhen8838: I don't need this test file. Delete it.

tests/analysis/test_isl_utility.py
zhen8838: Removed the direct normalize_dim test. The parser dimension test remains the end-to-end coverage for integer scalar HIR normalization.

src/tilefoundry/visitor_registry/op_cost.py
zhen8838: Yes, MatMul FLOPS remains 2 * batch * m * k * n. The fix for #117 only changes where the local dimensions come from: batch, m, and n now come consistently from the output shape, while k comes from lhs, avoiding double-counting M after a sharded axis expands the local rank.

docs/spec/parser.md
zhen8838: The rule visitor reached the same owning element through three parser situations. The generated table now merges rows by owner and rule, and lists all situations in one cell; the spec has been regenerated.

tests/parser/test_spec.py
zhen8838: Deleted tests/parser/test_spec.py as requested.

tests/models/qwen3_5_35b_a3b/model.py
zhen8838: The authoring tutorial contains no concat or stack calls, so there was no source cell to migrate. I scanned the shipped examples, which already use explicit sequences, and added the single-sequence contract to docs/tutorial/migrate.md.

src/tilefoundry/visitor_registry/op_cost.py
zhen8838: Follow-up in ada0146: MatMul now owns explicit a_layout Literal[MK, KM] and b_layout Literal[NK, KN], while the defaults remain the original MK/KN. Access relation, type inference, local-K cost recovery, evaluator, and target atom selection all use the same axis mapping; regression coverage includes all four layouts and a sharded KM/NK K projection.

PR #124, opened 2026-08-26
feat(parser): make authored expression gates work and show themselves

zhen8838:
## Why
- Three authored-expression gates were broken or invisible. A Python float
  scalar never adopted its peer's dtype, so `x + 1.0` on a bf16 tensor was
  rejected although hir §1.3 requires the authoring surface to adapt it. `@`
  was not a runtime operator at all, so `x @ b` could not be written. The
  generated grammar rendered `binary-op` as a placeholder, so the operator set
  the parser really accepts was unreadable.
- A refused call reported `runtime_expression: nested pattern failed`. A choice
  synthesized a failure for every alternative that returned `None`, so a report
  listed one line per alternative tried and the reason had to be dug back out
  of the pile.
- HIR carried per-shape CUDA MMA ops. A backend-bound construct in the
  reference side pulls HIR toward the imperative, and TIR already owns that
  surface.

## What
- `_constant` produces a dtype-less weak Constant for Python floats; it takes a
  typed floating-point peer's dtype before type inference. Unary minus keeps it
  weak. Python integers and tensor/tensor mismatches stay strict.
- `MatMulExpressionPattern` builds `MatMul` from `ast.MatMult` in runtime
  position only; the placement sugar `2 @ mesh.tile` is untouched.
- Matching gains a typed third outcome: a pattern that has established the node
  is its own refuses with a reason, every other combinator returns that refusal
  unchanged, and a choice records only the alternatives that claimed the node.
- Refusals name what is wrong with a call, a mesh, a `launch`, or a static dict
  where those sites previously returned `None`.
- The generated grammar expands the real operator set, adds
  `matmul-expression`, and states each loop iterator's argument count.
- `ir/hir/cuda/` is deleted along with its lowering refusal, cost evaluator,
  tests, and the isolated `mma_tile` fixture.

## Contract
- parser: `match` returns `AstMatch`, `None`, or a `MatchFailure`. A pattern
  MUST claim a node before refusing it with a reason; a pattern that has not
  claimed it returns `None`. The generated grammar and constraint table change
  with the patterns.
- hir: the CUDA matrix multiply-accumulate family is removed; `Mma` leaves the
  symmetric multi-input list. passes: the CUDA HIR MMA refusal section is
  removed. code-organization: MMA is target-owned under `ir/tir/cuda/nn/` only.
- TIR's `T.mma`, its atom, and CUDA codegen are unchanged.

## Risk
- The claim discipline is per-pattern. A pattern that claims a shape another
  alternative owns produces a misleading report; the corpus check is that a
  valid program yields no refusal at all, which currently holds.


docs/spec/parser.md
```diff
@@ -188,13 +192,18 @@ function              ::= 'def' name '(' signature ')' ('->' return-type)? ':' b
 | index_slice | subscript_index | TileWindowSliceBoundRule | A tile window cannot be used as a slice bound. | src/tilefoundry/parser/pattern_nodes.py |
 | layout | tensor_optional_slot | LayoutPositionRule | A layout must be legal for its parser position. | src/tilefoundry/parser/ast_pattern.py |
 | layout | tensor_optional_slot | LayoutShapeRule | A layout must have a valid non-boolean shape. | src/tilefoundry/parser/ast_pattern.py |
+| matmul_expression | expression, slice_endpoint, subscript_index | CallBindingRule | A call must bind its arguments into a Call tuple. | src/tilefoundry/parser/pattern_nodes.py |
```
zhen8838: Repeated 4 times

src/tilefoundry/ir/hir/math/binary.py
```diff
@@ -128,13 +128,31 @@ def _(call: "Call", ctx: "TypeInferContext") -> TensorType:
     lhs_ty = ctx.type_of(call.args[0])
     rhs_ty = ctx.type_of(call.args[1])
     if lhs_ty.dtype != rhs_ty.dtype:
+        number = next(
```
zhen8838: Why does this solve what problems?

src/tilefoundry/parser/ast_pattern.py
```diff
@@ -108,11 +108,61 @@ def _pattern_label(pattern: object) -> str:
     return type(pattern).__name__
 
 
+def _runtime_expression_detail(node: object) -> str:
```
zhen8838: Wrong design

tests/analysis/test_analysis_families.py
```diff
@@ -290,7 +290,7 @@ def test_the_placed_inventory_takes_the_whole_directory() -> None:
     analyses run; a count is not an answer, so this runs nothing.
     """
     roots = placed_fixture_roots()
-    assert len(roots) == 29
+    assert len(roots) == 28
```
zhen8838: What's the meaning of this kind of test? remove it

tests/parser/test_matmul_syntax.py
```diff
@@ -0,0 +1,109 @@
...
+        _HEADER
+        + """from tilefoundry.ir.types.shard import Mesh, Topology
+from tilefoundry.target import CudaTarget
+
+@module(
+    entry="placed_matmul",
+    target=CudaTarget("nvidia.h200_sxm"),
+    topologies=(Topology("cta", 2),),
+)
+class PlacedMatMul:
```
zhen8838: The test writing method is wrong, and then merge it into the test call.

tests/parser/test_runtime_expression_diagnostics.py
```diff
@@ -0,0 +1,75 @@
...
+            "runtime_expression: unsupported Dict syntax \"{'value': x}\"",
+        ),
+        (
+            "return tf.grouped_matmul(x, x) + x",
+            "runtime_expression: unsupported call 'tf.grouped_matmul' "
+            "(2 positional, no keywords)",
+        ),
+    ),
+)
+def test_unsupported_expression_names_its_shape(body: str, expected: str) -> None:
```
zhen8838: Delete this file and merge the previous test files.

tests/parser/test_scalar_dtype.py
```diff
@@ -0,0 +1,89 @@
...
+            + """@func
+def add_integer(x: Tensor[(8,), "bf16"]):
+    return x + 1
+"""
+        )
+
+
+def test_tensor_operands_do_not_promote() -> None:
+    with pytest.raises(VerifyError, match=r"bf16 vs f32.*never promoted"):
+        import_dsl(
```
zhen8838: Wrong test writing method

Issue #118, opened 2026-08-25
fix(analysis): terminate when an insert_slice offset is a product of a loop induction variable


zhen8838:
## Why
- `analyze` does not finish on a program whose `insert_slice` destination offset is a **product of a loop induction variable**. The same program with that one offset made literal finishes in 7 s.
- It is not a big program: one mesh, one row loop, one strip-group loop of 3 trips, 48 matmuls.

```
R0_literal              7.4 s   ideal-ns=139407
R1_weight_only         19.0 s   ideal-ns=418219      # sb only in the ConstTensor window
R2_store_only        >300.0 s   (timeout)            # sb only in the insert_slice offset
R3_both              >300.0 s   (timeout)
```

## What
- `sb = group * NS` with `group` a `range` induction variable.
- Used as a **window base** on a `ConstTensor` (`w[sb:sb + NS, ...]`) it costs 2.6x and completes.
- Used as an **`insert_slice` destination offset** (`insert_slice(acc, o, (sb, base, 0))`) it does not complete.
- `base = t + 0` in the *same* offset tuple is fine, and `sb = group + 0` in the same position is fine, so it is the product and the destination together, not symbolic offsets as such.

## Contract
- No behavioural contract: the analysis is correct when it terminates. What changes is whether a legal program can be analysed at all.
- Consequence for authors: a wide projection cannot be written as one stage looping over its strip groups; it has to be split into one stage per group so the offset stays literal. That is a real structural constraint imposed by an analysis cost.

## Risk
- Likely the polyhedral footprint of a write whose base is a non-affine-looking product; if so, the fix is to recognise `induction * constant` as affine rather than to widen the solver budget.
- `--memory` is required for `--roofline`/`--performance`, so there is no flag combination that avoids it.

## Reproducer

`tools/issues/analyze_symbolic_offset.py`, inlined — four cases from one template, differing only in `__WIN__` / `__STORE__`:

```python
from __future__ import annotations
from tilefoundry import func, module
from tilefoundry.dsl import ConstTensor, DimVar, Mesh, Tensor, tf
from tilefoundry.dsl.tf import *  # noqa: F401, F403
from tilefoundry.ir.types.shard import Topology
from tilefoundry.target import CudaTarget

H, GRID, HP, NHS = 2048, 128, 128, 16
NS, NM, RTG, NGROUP = 32, 4, 256, 3
NOUT = NGROUP * NS
S = DimVar("seq_len", 0, 8193)
[...]

PR #126, opened 2026-08-27
refactor(analysis): unify analysis structure and fix cross-scope and symbolic-offset analysis

zhen8838:
## Why
- #122 and #118 share one root cause: definition order, loop ownership, and affine
  offsets were each rebuilt per analysis family, and whichever copy was incomplete
  produced the failure.
- So the fix consolidates them behind one shared `StructuralMemo` and one shared
  affine entry point, rather than patching each family separately.

## What
- One structural memo and one loop-affine parser shared by every analysis family;
  `poly/`, `check.py`, and `performance.py` reorganised around them; two regressions added.

## Contract
- `AnalysisCallable` changes from five parameters to `(function, context)`; see
  `docs/spec/analysis.md`.
- `tilefoundry analyze` is bounded at 300s wall clock; on expiry it reports
  `analysis too complex` and exits nonzero. Currently fixed, not configurable; see
  `docs/spec/cli.md`.
- Analysis ownership moves to `poly/`, `performance.py`, a merged `check.py`, and the
  new `visitor.py` and `poly/affine.py`; see `docs/spec/code-organization.md`.

## Risk
- The CLI timeout is not configurable.
- An unbounded ISL parameter is still over-approximated by `_widest_allowed` rather
  than refused.


src/tilefoundry/analysis/check.py
```diff
@@ -892,4 +883,109 @@ def check_program(
     return derived
 
 
-__all__ = ["AnalysisCheckContext", "PerformanceInputChecker", "check_program"]
+def infer_authored_types(
```
zhen8838: Why is there another inference type here? Can‘t the inference type we provide use it?

src/tilefoundry/analysis/footprint.py
```diff
@@ -3,83 +3,21 @@
...
-class _RelationCase:
-    """One domain of a Call and the original operands modeled on that domain."""
-
-    domain: isl.set
-    relations: AccessRelations
-    local_types: tuple[object, ...]
-
-
 def _local_type(type_: object) -> object:
     """Narrow every Split axis while preserving the tensor's logical rank."""
```
zhen8838: Why do we need a local type? Didn‘t we already provide the local type?

src/tilefoundry/ir/core/__init__.py
```diff
@@ -9,7 +9,10 @@
     ExecutionDomainMetadata,
     IRMetadata,
     SourceSpanMetadata,
+    attach_metadata,
     binding_name,
+    describe_expr,
```
zhen8838, line 14: What is this for?

src/tilefoundry/ir/types/__init__.py
```diff
@@ -5,11 +5,14 @@
 from .dtype import BoolDType, DType, FloatDType, IntegerDType
 from .tensor_type import TensorType, TupleType, Type, UnitType
 from .utils import (
+    bytes_by_storage,
```
zhen8838, line 8: These three derived functions feel repetitive.

src/tilefoundry/ir/visitor.py
```diff
@@ -344,6 +348,84 @@ def default_visit(self, expr: Expr, ctx: Any) -> Expr:
 from tilefoundry.ir.hir.function import Function as HirFunction  # noqa: E402
 
 
+def expr_operands(expr: Expr) -> tuple[Expr, ...]:
```
zhen8838: This function is also very redundant.

src/tilefoundry/analysis/poly/__init__.py
```diff
@@ -0,0 +1,26 @@
+"""Public polyhedral analysis API."""
+
+from .access import (
+    AccessFootprint,
```
zhen8838, line 4: We have accessed and looped in Analysis, and there is another set here. If no one uses poly now, just remove it all.

src/tilefoundry/analysis/check.py
zhen8838: Resolved in M2 commit 5035d2b: check_program now clones first, then performs exactly one inference_type(derived.body, ctx). The earlier authored inference pass was removed; grep shows one call site.

src/tilefoundry/analysis/poly/__init__.py
zhen8838: Tracked as issue #132: schedule still has four importers of the poly access concepts (pipeline/program.py, kernel_schedule.py, render.py, and pipeline/problem.py), so removal is not safe. Converging analysis/poly and analysis/scope is outside this PR because it requires a schedule contract change.

src/tilefoundry/analysis/footprint.py
zhen8838: Resolved across M1 and D11: local_type_of is the single implementation in ir/types/utils.py. The footprint copy was moved in M1 and the poly copy was removed in D11; grep shows no remaining _local_type implementation or import.

src/tilefoundry/ir/core/__init__.py
zhen8838: Resolved in M4 commit 7167b43: the redundant expr_operands helper was removed and the shared public expr_children API is now used by all callers.

src/tilefoundry/ir/types/__init__.py
zhen8838: Resolved in M4 commit 7167b43: numel and tensor_bytes now fold through tensor_types, with _leaf_numel preserving the existing dynamic and negative extent diagnostics.

src/tilefoundry/ir/visitor.py
zhen8838: Resolved in M4 commit 7167b43: expr_operands was removed; expr_children is the shared public traversal API used by the dataflow and inspection callers.

PR #129, opened 2026-08-27
refactor(ir): unify function call typing and call edges

zhen8838:
## Summary

- keep HIR Function call targets as authored templates and move call typing into `TypeInferVisitor`
- replace elaboration state with scope-local type bindings and traversal-wide result-type memoization
- standardize expression traversal around `ExprFunctor`, `ExprVisitor`, `ExprWalker`, `ExprCollector`, and `ExprCloner`
- make dimension instantiation, evaluation context, and one-shot type inference use the standard visitor contracts
- remove obsolete call-edge helpers and let each preorder/postorder consumer use `collect_exprs` directly
- retype flattened analysis views after inlining and remove the obsolete per-call materialization test

## Contract

`ExprCloner` memoizes by expression identity. When two inputs share one expression, their cloned outputs now share one result as well; the prior `ExprMutator` test assertion was intentionally reversed to lock in this contract.

Public additions: `ExprCloner`, `ExprCollector`, `collect_exprs`, `inference_type`, `DimensionInstantiator`, `InstantiateContext`, `instantiate_dimensions`, and `EvaluateContext`.

Public removals: `ExprMutator`, `calls_in`, `calls_in_expr`, `ChildModuleResolver`, `EvalContext`, and `FunctionEvalContext`.

## Verification

- authoritative run: `RUN=visitor-m3-full-20260828-01 tilefoundry-verify source`
- commit: `e6bfb5e`, identical to this PR's head; M4 was cut without a new commit, so this run covers the final tree
- result: 910 tests, 906 passed, 4 existing skips, 0 failures
- manifest: `test_results/visitor-m3-full-20260828-01/manifest.md`

## Risk

D41 adds a type-inference traversal after analysis builds its flattened view. Across the current full corpus (54 modules, 888 derived nodes), this traversal changes zero Type values: no callee is currently called from sites carrying different layouts. It closes a latent stale-type path rather than fixing an observable current failure.

Under D8 this refactor adds no new tests, so no regression test fails if the D41 retype step is later removed. Structural review and the existing full suite are the present guard; this is the residual risk.

Call targets now remain shared authored templates instead of per-site Function copies. The one test that required per-site identity was removed by explicit author decision; the retained attached-copy roundtrip test continues to assert module-owned targets stay distinct.


docs/spec/core-ir.md
```diff
@@ -91,6 +91,15 @@ invocation. Owning a `Function`, attaching a child `Module`, or declaring a
 Python-to-HIR entry and HIR-to-HIR device calls are governed by
 [hir §1.1](./hir.md#11-function).
 
+Call-edge discovery is shared by every consumer of a Function body:
+
+- `calls_in(function)` returns the HIR Function calls in the selected body;
```
zhen8838: This thing is beyond the scope of my design. It is necessary to explain the reasons that must be added.

docs/spec/visitor-registry.md
```diff
@@ -226,16 +211,23 @@ nothing of that kind rather than guessing.
...
-    it MUST NOT carry a Module, scope, reading, or parser metadata.
-  - `CallFeedProvider` is context-owned: Parser, Type Inference, and function-level
-    Evaluator provide their own value type and ownership rules. HIR consumes
-    `TypeInferContext.build_call_feed()` and does not own a resolver.
-  - `type_of` is a walk-local cache only — it holds no dispatch rule of its
-    own. A cache miss delegates to `TypeInferVisitor(self).visit(expr)`
-    (below), whose `visit_Call` is what consults
-    `typeinfer_registry.lookup(type(target))`; an unregistered Op call
-    routes through `ctx.error`.
+  - `child_resolver` is parser-owned child-module resolution state before a
```
zhen8838: Why do you need this?

docs/spec/visitor-registry.md
```diff
@@ -258,38 +250,39 @@ Visitor:
...
-    def visit_Call(self, call: Call): ...            # typeinfer_registry.lookup(type(call.target))
-    def visit_Tuple(self, tup: Tuple): ...            # structural: TupleType over each element's type
-    def visit_GridRegionExpr(self, grid): ...         # carry/body — hir §1.2
-    def visit_ShapeOf(self, shape_of: ShapeOf) -> Type: ...
+    def __init__(self, ctx: TypeInferContext, *, memo=None): ...
+    def visit_leaf_Var(self, var: Var, operands): ...
+    def visit_leaf_Constant(self, c: Constant, operands): ...
+    def visit_leaf_Call(self, call: Call, arg_types): ...
+    def visit_leaf_Tuple(self, tup: Tuple, field_types): ...
+    def visit_operands_GridRegionExpr(self, grid): ...
```
zhen8838: Why Visit Operands instead of Visit?

src/tilefoundry/evaluator/context.py
```diff
@@ -4,8 +4,6 @@
 from dataclasses import dataclass
 from typing import Any
 
-from tilefoundry.visitor_registry.contexts import CallFeed
-
 
 @dataclass(frozen=True)
 class EvalContext:
```
zhen8838: These two contexts need to be combined into one.

src/tilefoundry/ir/core/module.py
```diff
@@ -71,27 +71,42 @@ def child_module_of(root: "Module", caller: object, callee: object) -> "Module |
     return called if any(child is called for child in owner.modules) else None
 
 
-def _calls_in(function) -> tuple:
-    """Every ``Call`` on a Function that the body of *function* makes.
-
-    Its own body and nothing else: a variant this dispatch did not select is
-    not running, and a converter runs offline rather than here.
-    """
+def calls_in_expr(
```
zhen8838: We need to discuss where to put this function.

src/tilefoundry/ir/hir/specialize.py
```diff
@@ -177,10 +193,212 @@ def specialize_function(
...
-        chosen, bound, ctx if ctx is not None else TypeInferContext(), dims=dims
+    return _rebuild_at(
+        chosen,
+        bound,
+        ctx if ctx is not None else TypeInferContext(),
+        dims,
+    )
+
+
+def _rebuild_at(
```
zhen8838: I need to explain why this is needed.

src/tilefoundry/ir/visitor.py
```diff
@@ -232,22 +234,43 @@ def __init__(
         and return a result belonging to a different expression.
         """
         super().__init__()
+        self.ctx = ctx
```
zhen8838: Ctx cannot be used as the default parameter of the visitor base class.

src/tilefoundry/visitor_registry/contexts.py
```diff
@@ -27,29 +27,6 @@
...
-
-class CallFeedProvider(Protocol[T]):
-    """Context-owned call binding and callee-scope construction."""
-
-    def build_call_feed(self, callee: object, supplied: tuple[T, ...]) -> CallFeed[T]: ...
-
-    def scope_for(self, callee: object) -> FunctionScope | None: ...
-
 
 def _constant_type(value: object) -> TensorType:
```
zhen8838: This function should not be needed, right?

src/tilefoundry/script.py
```diff
@@ -314,14 +314,10 @@ def parse(self):
         )
         if self.dialect == "hir":
             if self.role is FunctionRole.VARIANT:
-                object.__setattr__(self.parsed, DISPLAY_NAME, self.binding_name)
-                object.__setattr__(self.parsed, "name", base.name)
+                setattr(self.parsed, DISPLAY_NAME, self.binding_name)
```
zhen8838, line 317: Why do we still need to use setattr here?

tests/ir/types/test_mma_fragment_layouts.py
```diff
@@ -82,7 +83,7 @@ def _assert_reshard_typeinfer_ok(
 
     call = Call(type=src_ty, target=op, args=(src,))
     ctx = TypeInferContext()
-    out_ty = ctx.type_of(call)
+    out_ty = TypeInferVisitor(ctx).visit(call)
```
zhen8838: Add a function of inference type to type utils

PR #130, opened 2026-08-28
fix(parser): enforce fused boundary annotations

zhen8838:
## Why
- Fused HIR boundaries dropped the shard/storage constraints authors wrote on signatures, kept return annotations that contradicted the body, and accepted a storage the target does not have.

## What
- Consolidate call-boundary validation, let a function declare its mesh on the decorator, make the inferred return authoritative, connect the parser's storage check to the target's memory facts, and unify mismatch diagnostics.

## Contract
- `types_compatible` compares declared fields recursively and is the single implementation of that rule: `None` is undecided and binds from the argument; `UMAT` is undecided storage. Everything stated is a constraint.
- `@func(mesh=Mesh(...))` is new: it names the function's outer execution domain, usable in the signature. `Mesh.topologies` accepts `str` level names, resolved against the owning `@module` at parse time.
- HIR return annotations no longer enter the IR; variants keep the base signature. Dispatch prototypes still require one.
- H200 no longer exposes a `tmem` level. HIR and TIR annotations, in signature and body, validate storage against the target's levels.
- `as_script` falls back to verbose `ShardLayout(...)` for a compound placed extent with explicit strides, which the compact sugar cannot express.
- `hir.md`, `parser.md`, and `target.md` are updated to match.

## Risk
- G3 stays open for dispatch variants: a variant's body type may differ from the base signature callers are given (`prefill_decode_attention`). Recorded as a follow-up.


src/tilefoundry/evaluator/interpreter.py
```diff
@@ -20,7 +20,7 @@
 )
 from tilefoundry.ir.core import Call, Constant, Tuple, Var
 from tilefoundry.ir.core.pattern import locate_dim_var
-from tilefoundry.ir.hir.function import Function
+from tilefoundry.ir.hir.function import Function, expected_arity_if_mismatched
```
zhen8838: This function should not be exported here, and the function of the function is doubtful.

src/tilefoundry/ir/types/utils.py
```diff
@@ -20,20 +20,51 @@
...
+            return "shape"
+        if declared.dtype != actual.dtype:
+            return "dtype"
+        if declared.storage is not StorageKind.UMAT and declared.storage != actual.storage:
+            return "storage"
+        return _layout_mismatch_field(declared.layout, actual.layout)
+    return None if actual == declared else "type"
+
+
+def types_compatible(declared: Type, actual: Type) -> bool:
```
zhen8838: They should all follow the compatible naming method.

src/tilefoundry/evaluator/interpreter.py
zhen8838: expected_arity_if_mismatched has been deleted; the two callers inline the original count logic, and the arity messages are unchanged.

src/tilefoundry/ir/types/utils.py
zhen8838: type_mismatch_field and _layout_mismatch_field have been deleted; types_compatible is again the sole boolean recursive implementation, and errors print the full declared and actual types.

PR #141, opened 2026-08-30
fix(diagnostics): preserve authored failure context

zhen8838:
## Why
- Parser-generated and synthetic calls could lose source provenance, while evaluator failures leaked backend-only errors.

## What
- Attach immutable authored spans to every parser-created Call, including detached tuple projections, using physical source columns.
- Wrap Op dispatch failures with their source, binding, and operation; fail closed for Local over Split layouts.
- Add diagnostic fixtures, installed CLI smoke coverage, and update affected evaluator expectations.

## Contract
- Parser-authored Calls retain their first precise span; tuple projections use their corresponding assignment targets.
- Op handler and missing-handler failures surface as EvalError prefixed by describe_expr(call).

## Risk
- Local evaluation for Split layouts remains intentionally unsupported until the mesh evaluator implements that path.


docs/spec/parser.md
```diff
@@ -22,6 +22,22 @@ optional base/key for one parse. `FunctionRole` is `ROOT`, `VARIANT`, or `CONVER
 `ParseError` is the single authored-source diagnostic type and includes source location and
 recursive parse situation. These are the only public parser symbols.
 
+### 1.1 Authored provenance
```
zhen8838: This section is too awkward. I understand that this is a behavioral constraint of a parser, a constraint on the state of the final product. So, can it be placed in the place of the Parser function above? It should be placed in the form of bullet. This is what information the call node will bring, and what kind of additional processing will the tuple get item have?

src/tilefoundry/parser/parser_visitor.py
```diff
@@ -35,25 +36,49 @@ def visit_function(self, node: ast.FunctionDef) -> Any:
         return self.visit(node)
 
 
-def _extract_function_def(fn: FunctionType) -> ast.FunctionDef:
+def _dedented_source(source_lines: list[str]) -> tuple[str, int]:
```
zhen8838: Does it need to be so complicated to get a dedented source? Isn‘t it on ast?

tests/analysis/test_tuple_projection_diagnostics.py
```diff
@@ -0,0 +1,52 @@
...
+
+from __future__ import annotations
+
+import ast
+from pathlib import Path
+
+import pytest
+
+from tests.fixtures.diagnostics.tuple_projection_diagnostic import TupleProjectionDiagnostic
+from tilefoundry.analysis.api import analyze
```
zhen8838: This test should be put in installed.

src/tilefoundry/parser/parser_visitor.py
zhen8838: Simplified it: parse normally; only indented inspected source is parsed unchanged under a temporary `if` suite. AST keeps the authored columns, so there is no dedent or column-offset bookkeeping.

tests/analysis/test_tuple_projection_diagnostics.py
zhen8838: Moved it to the installed `tilefoundry analyze --performance` smoke. The fixture remains external to the placed-success inventory.

docs/spec/parser.md
zhen8838: Moved the provenance contract under `parse_function` as two bullets: the generic Call span invariant and the TupleGetItem exception.

PR #142, opened 2026-08-31
fix(parser): validate HIR return annotations

zhen8838:
## Why
- HIR return annotations were parsed but did not validate the inferred body result, allowing real dispatch variants to silently disagree with their declared return contract.

## What
- Validate annotated HIR function bodies and dispatch variant bodies with `types_compatible`.
- Require a return annotation on `pass` dispatch prototypes and support recursive `tuple[...]` return annotations.
- Add a diagnostic regression test and make the decode fixture explicitly materialize its declared GMEM result.
- Remove unrelated ordinary-function annotations whose bodies intentionally remain SMEM or UMAT.

## Contract
- Omitted `Tensor[...]` storage remains GMEM; it is not a wildcard.
- Ordinary functions keep `body.type` as their IR return type after validation. Dispatch variants keep the exact prototype return type.

## Risk
- Existing authored return annotations now need to describe the body result. The full source suite passes after updating the affected fixtures and tests.

### DIFF
--- a/docs/spec/hir.md
+++ b/docs/spec/hir.md
@@ -116,6 +116,16 @@ declares, but it MUST NOT create a level or change one's extent.
 `return_type`. The projection is fixed at construction and stays
 consistent across construction sites.
 
+**Parser return contract.** At the DSL boundary, a body with an authored return
+annotation MUST satisfy `types_compatible(annotation, body.type)`. A normal
+function keeps its inferred `body.type` as `return_type` after that check. A
+`pass` dispatch prototype MUST declare its return type; it becomes the base
+return type and every variant body MUST satisfy it before the variant receives
+that exact base `return_type`. `Tensor[...]` without a storage slot is a GMEM
+annotation, including here: it is not an unconstrained storage spelling. The
+parser accepts `tuple[...]` and applies the same compatibility relation
+recursively to its fields. See [parser §3.1](./parser.md#31-hir-return-contracts).
+
 **Call typing — visitor-scoped inference.** A `Call` keeps its authored
 `Function` template as `target`. Its result type is inferred by seeding a new
 visitor memo with the actual argument types bound to the callee's formal
--- a/docs/spec/hir.md
+++ b/docs/spec/hir.md
@@ -111,20 +111,15 @@ parser-lexical mesh binding; `ShardLayout.mesh` MUST point at an active
 binding on the lexical path. A `Mesh` MAY map fewer levels than the domain
 declares, but it MUST NOT create a level or change one's extent.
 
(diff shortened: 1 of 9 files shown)


docs/spec/hir.md
```diff
@@ -116,6 +116,16 @@ declares, but it MUST NOT create a level or change one's extent.
 `return_type`. The projection is fixed at construction and stays
 consistent across construction sites.
 
+**Parser return contract.** At the DSL boundary, a body with an authored return
```
zhen8838: `Function.return_type` should be written here as the beginning. Then write the parse-related in the parser, and the details will not be introduced here.

docs/spec/parser.md
```diff
@@ -219,6 +223,27 @@ function              ::= 'def' name '(' signature ')' ('->' return-type)? ':' b
 | Pattern Visitor | Traverses the same graph to render this section's generated grammar and constraints. |
 | Refusal | Carries the reason from the pattern that claimed a node and then refused it, so a report names a cause rather than the absence of a match. |
 
+### 3.1 HIR Return Contracts
```
zhen8838: A contract of parser behavior should be added after parse_function in Chapter 1, and then a section in it introduces the return contract. It turns out that these bullets are also in the form of sections:
```text
- Every parser-authored `Call` reachable from a Function body carries `SourceSpanMetadata` for
  the AST expression that constructed it. A parent match fills only Calls without a span, so it
```

zhen8838: Added HIR Return Contracts as parser section 1.1 immediately after parse_function and removed the duplicate from section 3.

docs/spec/hir.md
zhen8838: Moved parser details out of HIR and retained only the IR-level Function.return_type definition there.

docs/spec/parser.md
```diff
@@ -17,6 +17,27 @@ def parse_function(
...
+type remains the exact base type so all variants share one dispatch signature.
+
+`Tensor[...]` without a storage slot constructs `storage=GMEM`, including in a
+return annotation. It is not an unspecified-storage spelling. Consequently an
+SMEM body under `-> Tensor[...]` must explicitly return a GMEM result; the
+parser reports both the annotation and inferred body type, with the authored
+function location, when they are incompatible. `tuple[...]` annotations are
+accepted and apply this same compatibility rule recursively to every field.
+
 - Every parser-authored `Call` reachable from a Function body carries `SourceSpanMetadata` for
```
zhen8838: You need to change the remaining bullets into small chapters.

zhen8838: Split the remaining public parser contracts into sections 1.2 through 1.4.

PR #144, opened 2026-08-31
refactor(check): layer check into selection / inputs / execution / comparison

zhen8838:
## Why
- `check` eagerly materialized every declared weight before selecting a leaf, so checking one runnable function could allocate the union of unrelated sibling weights and fail before execution.
- Selection, input acquisition, execution, and comparison were coupled behind overlapping flags and executable IR wrapper objects.

## What
- Split `check` into explicit selection, activation/weight acquisition, concrete execution, and report comparison layers.
- Added lazy runtime resources with weak safetensors read caching and strongly memoized deterministic random draws shared across subtree views.
- Moved semantic execution into the evaluator through a transient `Reading`, preserving lazy child resources and host-side orchestration.
- Restored reports for input provenance, declared and actual dtypes, selected variants, pinned dimensions, reference norms, warnings, and matching text/JSON facts.

## Contract
- Replace `--input` / `--ckpt` with independent `--inputs random|files:A,B` and `--weights random|ckpt:DIR` axes.
- `LoadedModule` is frozen data containing `module`, `resource`, and child `modules`; it neither owns materialized tensors nor executes functions.
- `Module.<fn>` resolves to the `Function` / `PrimFunction` IR node itself, as specified by core-ir section 1.1.
- Remove `Module.forward` and `Module.__call__`; semantic execution goes through the evaluator only.
- `evaluate` accepts a `LoadedModule`, supports explicit `function=`, and takes execution placement from caller-provided `device=`.
- Export `evaluator.reading(loaded, device=...)` as the public host-orchestration view.

## Risk
- Ten files under `examples/` still use the old executable-loaded-module API. They are outside pytest and require a separate migration plan.
- One-to-many `load_group` aliases and the two-level runtime twin base-class structure are intentionally unchanged.


docs/spec/evaluator.md
```diff
@@ -26,21 +26,35 @@ flowchart TB
...
-    fn_or_call: "Function | Call",
+    fn_or_call: "Function | Call | LoadedModule",
     *inputs: "torch.Tensor",
     backend: str = "torch",
     device: str | None = None,
+    function: str | None = None,
 ) -> "torch.Tensor | tuple[torch.Tensor, ...]":
     ...
+
+def reading(loaded: "LoadedModule", *, device: str) -> "Reading":
```
zhen8838: What does `reading` do?

docs/spec/evaluator.md
```diff
@@ -26,21 +26,35 @@ flowchart TB
 
 ```python
 def evaluate(
-    fn_or_call: "Function | Call",
+    fn_or_call: "Function | Call | LoadedModule",
     *inputs: "torch.Tensor",
     backend: str = "torch",
     device: str | None = None,
+    function: str | None = None,
```
zhen8838: It should still be a data structure, packaged as fn_or_call so that it can be selected, instead of adding function as a parameter.

src/tilefoundry/cli/check.py
```diff
@@ -404,19 +413,19 @@ def _tensor_leaves(value):
...
 
 def _tensor_structure(value):
-    """The JSON-safe shape tree that an activation file supplied."""
+    """Build the JSON-safe shape tree supplied by an activation file."""
     if isinstance(value, torch.Tensor):
         return {"dtype": str(value.dtype), "shape": list(value.shape)}
     return [_tensor_structure(item) for item in value]
 
 
 def _dtype_names(values: Sequence[Any]) -> list[str]:
```
zhen8838: Isn't this kind of function ready-made in the warehouse? Do you still need to redefine?

src/tilefoundry/cli/check.py
```diff
@@ -466,17 +454,23 @@ def _variant(function: Function, dims: dict[str, int]) -> dict[str, Any] | None:
 
 
 def _pin(function: Function, stated: dict[str, int]) -> tuple[dict[str, int], list[dict[str, Any]]]:
```
zhen8838: You shouldn't need a pin. You can choose it by yourself in the evaluator.

tests/dsl/test_module_decorator.py
```diff
@@ -187,18 +187,22 @@ def scale(x: Tensor[(2,), "f32"], w: ConstTensor[(2,), "f32"]):
     ones = torch.ones(2, dtype=torch.float32)
     weight = torch.full((2,), 3.0)
 
-    assert _Weighted.scale(ones, weight).float().cpu().tolist() == [3.0, 3.0]
+    from tilefoundry.evaluator import evaluate  # noqa: PLC0415
```
zhen8838: It should not be imported internally.

src/tilefoundry/cli/check.py
```diff
@@ -675,6 +469,51 @@ def _shown_files(files: Sequence[dict[str, Any]]) -> str:
     return "; files " + "; ".join(descriptions)
 
 
+class _TrackedResource:
```
zhen8838: shouldn't add this.

src/tilefoundry/evaluator/interpreter.py
```diff
@@ -325,7 +395,17 @@ def evaluate(fn_or_call, *inputs, backend: str = "torch", device: str | None = N
         raise EvalError(f"evaluator: unsupported backend {backend!r}")
     device = device or _default_device()
 
-    if isinstance(fn_or_call, Function):
+    from tilefoundry.ir.core.module import LoadedModule, _refuse_bare_call  # noqa: PLC0415
```
zhen8838: 1. What is _refuse_bare_call? Two. Why do you need to judge the loadedmodule specifically?

Issue #149, opened 2026-09-02
fix(parser): give a loop the source span its diagnostics need


zhen8838:
## Why

A loop carries no source identity, so no diagnostic about a loop can say where it
is. `attach_authored_metadata` attaches a span to a `Call` and to nothing else
(`src/tilefoundry/parser/ast_pattern.py:201-202`):

```python
def attach_authored_metadata(value: object, node: ast.AST, context: "MatchContext") -> object:
    """Attach *node*'s source identity to each unmarked Call it constructs."""
    if not isinstance(value, Call):
        return value
```

`GridRegionExpr` is not a `Call`, and neither is `Var`. Measured on a 21-line
program with one loop: 4 of 6 `Call`/`Var` nodes carry `SourceSpanMetadata`, and
the three that do not are exactly the loop, its `induction_var`, and its `extent`.
`diagnostic_location` returns `None` for all three.

This is what stopped #127 from pointing anywhere. That refusal now names the
variables instead:

```
tilefoundry: error: loop '_step' has a trip count the program computes at run
time from 'reps', so no per-occurrence total can be scaled by it; bind the
extent to a literal, or state it as an open dimension
```

Readable, but the reader still has to find `_step` themselves. Both refusals in
`_domain_for` (`src/tilefoundry/analysis/scope.py`) are in this position, and so
is any future one about a loop.

## What

Give a loop the same source identity a call has. The parser knows the `ast.For`
node's position at the point it builds the region.

## Contract

`describe_expr` and `diagnostic_location` (`src/tilefoundry/ir/core/metadata.py:65`
and `:77`) start answering for loops instead of returning `None`.

## Risk

**Scope needs settling before anyone starts, because `Var` and `GridRegionExpr`
are not the same decision.**

Attaching to the loop region alone is contained: it buys the diagnostics above
and changes no report.

Attaching to `Var` also moves report labels. `value_label`
(`src/tilefoundry/ir/core/metadata.py`) reads a span when there is one, so a
loop's carried argument would go from `acc` to `acc:19`, and a parameter from
`k_cache` to `k_cache:12`. Two consequences:

- `tests/installed/models/smoke_qwen3_1_7b.py:73` asserts
  `{"k_cache", "v_cache"} <= zero.keys()`, which holds today precisely because a
  parameter has no span. It would need to change.
- `value_labels`' repeated-name suffix exists for the case where two carried
  arguments share a name and neither has a span
  (`('acc', 'rmem')` twice, measured on
[...]

Issue #140, opened 2026-08-30
[BUG] analyze --memory times out on large-dimension reshape/transpose chains (ISL count_val is O(N))


KaiyuZhang001:
Describe the bug
Running tilefoundry analyze with the --memory flag on a function containing reshape/transpose operations over large-dimension tensors (e.g. N=961,600) takes over 120 seconds and times out. The same function analyzed with --compute-cost alone completes in ~2 seconds.
Root cause
The memory analysis calls reached_elements() in access_relation.py, which invokes isl.Set.count_val() to count the number of points in an integer set:
# tilefoundry/visitor_registry/access_relation.py:803
return int(str(image.coalesce().count_val()))
ISL's count_val() enumerates all points in the set, making it O(dim_value). For a tensor with dimension N=961,600, a single reshape triggers ~25s of ISL computation. A chain of 5 reshape/transpose ops exceeds 120s.
The call chain is:
analyze_memory → _record_movement → call_traffic → _movement → reached_elements → isl.count_val()
Reproduction
Minimal repro file (repro.py):
from __future__ import annotations
from tilefoundry import func, module
from tilefoundry.dsl import Tensor, tf
from tilefoundry.dsl.tf import *  # noqa: F401, F403
from tilefoundry.ir.types.shard import Topology
from tilefoundry.target import CudaTarget

B = 1536; H = 5; HD = 128; QKV = 768; L = 470; N = B * L 

@module(target=CudaTarget("nvidia.h100_sxm"), topologies=(Topology("cta", 132),))
class FullChain:
    @func
    def forward(q: Tensor[(N, QKV), "bf16"]):
        q4 = tf.reshape(q, new_shape=(B, L, H, HD))
        q4 = tf.transpose(q4, perm=(0, 2, 1, 3))
        qf = tf.reshape(q4, new_shape=(B * H, L, HD))
        o3 = tf.reshape(qf, new_shape=(B, H, L, HD))
        o3 = tf.transpose(o3, perm=(0, 2, 1, 3))
        return tf.reshape(o3, new_shape=(N, QKV))
Steps to reproduce:
# Fast — compute-cost only, no ISL (~2s):
python -m tilefoundry.cli analyze repro.py:FullChain.forward /tmp/out.txt --compute-cost --topology cta

# Slow — memory analysis triggers ISL count_val, >120s timeout:
python -m tilefoundry.cli analyze repro.py:FullChain.forward /tmp/out.txt --memory --topology cta
Expected behavior
[...]

Issue #133, opened 2026-08-29
fix(analysis): make loop-footprint reports reproducible across runs


zhen8838:
## Why

`--memory` reports are not reproducible: analysing the same program twice
prints a different `loop-footprint` line. The numbers are the same, but the
order of the entries and the label of an unnamed buffer both change from one
process to the next.

Observed on `55ad440` (and every commit around it), analysing
`test_results/m4-issue122-fix-r3/issue122.py` twice with
`--performance --memory --dim seq_len=1024`:

```
run 1   loop-footprint footprints=v3@smem:4294967296/549755813888/...
run 2   loop-footprint footprints=<buffer 136876484903760>@gmem:8796093022208/...
```

Even which entry comes first changes.

## What

`Scope.footprint` in `src/tilefoundry/analysis/scope.py` lets `id()` reach the
output in two places:

```python
key = (id(access.buffer), str(getattr(access.buffer.type, "storage", "unknown")))
...
for (buffer_id, level), values in sorted(rows.items()):      # sorts on the address
    BufferFootprint(buffer=binding_name(values[0]) or f"<buffer {buffer_id}>")   # prints it
```

Using `id()` to *group* is right -- two structurally equal buffers are distinct
allocations. Using it to *order* and to *name* is not: an address is whatever
the allocator handed out this run.

## Contract

`bytes`, `device_bytes` and `repeated_bytes` are unaffected: they are summed per
key, and addition does not depend on order. No consumer indexes into
`footprints` -- `inspection/values.py:276`, `analysis/memory.py:418`
(`cache_pressure`) and `tests/analysis/test_analyze_at_a_size.py:235` all
iterate or filter, so capacity verdicts do not change either. This is a
reproducibility defect, not a wrong number.

## Risk

It defeats the one thing analysis reports are for: comparing two runs. A reader
diffing a report against a stored baseline sees changes that are not there, and
`<buffer 136876484903760>` points at nothing they can find in their program.

A fix would order by something stable (binding name, then level, then the
buffer's definition position) and name an unnamed buffer by its definition
order rather than its address.

zhen8838:
Fixed by #150, landed on main as 18edc35.

`loop-footprint` no longer reads `id(buffer)` for the label or the sort order. A label is now the value's name plus the authored line that produced it, from `value_labels`, shared with the value lifetimes that `docs/spec/analysis.md:390` already said footprint labels must match. Grouping stays by buffer identity, which was the part that was right.

[...]

Issue #127, opened 2026-08-27
fix(analysis): distinguish an unknown trip count from a single trip


zhen8838:
`analyze` reports a loop whose trip count is computed at run time as if it ran
once. No warning, exit 0, report written.

`--dim` does not help: the bound is not an open dimension, it is a value the
program computes.

### The number is wrong by exactly the trip count

Same program, only the loop bound changes:

```
for _step in range(reps):   # reps is a runtime i64 tensor
  # compute-cost flops=f32:16896@128
  # traffic      traffic=gmem:r67584/w67584@r512/w512

for _step in range(8):
  # compute-cost flops=f32:135168@1024
  # traffic      traffic=gmem:r540672/w540672@r4096/w4096
```

8x, silently. `memory` does mark its own reading
(`loop-footprint status=lower-bound`), but `compute-cost` and `traffic` report
the one-trip figure with no marker at all, and `roofline`/`performance` build on
those.

### Where it comes from

`analysis/walk.py:40` collapses "unknown" into "one":

```python
def loop_trip_count(loop: GridRegionExpr) -> int:
    start, extent, step = loop.start, loop.extent, loop.step
    if not all(isinstance(value, int) for value in (start, extent, step)):
        return 1
```

The docstring argues one is better than a guess, and for a `DimVar` extent that
holds -- `check.py:556 _require_concrete_geometry` refuses an unbound dimension,
so that path never reaches here:

```
tilefoundry: error: analyze needs one EXTENT for every open dimension: seq_len
is declared as [0, 8193); bind it with --dim seq_len=EXTENT
```

But a *computed* extent is neither a topology extent nor an open dimension, so
neither `_require_concrete_geometry` nor `is_concrete(function)` rejects it. It
reaches `loop_trip_count`, becomes 1, and every multiplicity-aware total is
scaled by 1 instead.

The distinction the code needs and does not have is "one trip" versus "unknown
trip count". Today they are the same `int`.

### Reproducer

```python
from __future__ import annotations
from tilefoundry import func, module
from tilefoundry.dsl import Mesh, Tensor, Topology, tf
from tilefoundry.target import CudaTarget

_H200 = CudaTarget("nvidia.h200_sxm")
CTAS = 132
N = CTAS * 128


@module(entry="kernel", target=_H200, topologies=(Topology("cta", CTAS),))
class DynTrip:
    @func
[...]

PR #150, opened 2026-09-02
fix(analysis): make the analyze report reproducible, readable and fast; fix the placement sugar

zhen8838:
## Why

- Four defects in what the authoring surface and `analyze` hand back. #133 printed
  a different report for the same program every run: `loop-footprint` read
  `id(buffer)` for both the label and the sort order. #127 refused a
  runtime-computed trip count with an 1800-character `GridRegionExpr` repr that
  never said which loop. #140 (reported from outside) spent 38.7 s on `--memory`
  where `--compute-cost` answered in 1.7 s, counting each access image by
  enumerating its points. #120 made `X @ mesh.axis` mean one thing in a reshard
  layout and another in `Tensor[...]`, so a `tf.zeros` accumulator could not meet
  anything a resharded op produced.

## What

- One shared label derivation, `value_labels`, for both the loop footprint and the
  value lifetimes `docs/spec/analysis.md:390` already said must match: a value's
  name plus the authored line that produced it. Uniqueness is settled over the
  group, because SSA renaming covers `Call` but not `Var`.
- Both refusals in `_domain_for` name the induction variable and the value the
  bound comes from.
- `point_count` reads a box's cardinality off its bounds instead of enumerating.
  The four other counting sites share it, so trip counts and one-pass extents get
  the same fast path, not only the reported site.
- A Tensor shape slot keeps its extents as written. `PlacedLayoutPattern` answers
  the layout slot, whose only question is where a value goes; the shape slot has
  no operand to read a shape from and used to recover one from the layout, whose
  shape is the divided one. `PlacedShapePattern` keeps both answers side by side.

```
analyze --memory     38.7s -> 1.76s     (--compute-cost 1.71s, unchanged)
loop-footprint       3 orderings -> byte-identical across processes
trip-count refusal   1800+ chars -> 187
zeros + resharded    broadcast error -> both shape=(1,16,8192)
```

## Contract

- `docs/spec/analysis.md:390`/`:397`: a label is a name plus its source line,
  shared by `BufferFootprint.buffer` and `ValueLifetime.binding`; grouping stays
  by buffer identity and no label reads an address.
- A layout may be stated once, not twice: `Tensor[(8 @ m.b, 8), "f32", (8 @ m.b,
  8)]` is refused rather than one silently replacing the other. All 27 authored
  uses of shape-slot sugar still parse, 18 in the shipped nemotron example.
- No analysis number moves; the #140 report is byte-for-byte identical.
  `authoring.md` re-rendered, since `smoke_tutorial.py` reproduces it exactly.
- No new tests. Existing items gained conditions:
  `test_analyze_at_a_size.py`, `test_analysis_families.py`, `smoke_analyze.py`.

## Risk

- A loop still cannot say where it is: spans attach to a `Call` and nothing else,
  so `GridRegionExpr` and its extent carry none. Filed as #149, with why a loop
  and a parameter are different decisions.
- Same-arity branches cannot be told apart once chosen, because a child is not
  parsed until construct; the layout branch therefore carries a lookahead over
  its third slot and is tried first.
- #120's `matmul`-seeded accumulator workaround can now be undone; not done here.

Verified: `source` 922 passed / 4 skipped, `installed` 168 passed, 0 failures.


src/tilefoundry/parser/pattern_nodes.py
```diff
@@ -738,13 +754,19 @@ class TensorOptionalSlotPattern(ElementPattern):
...
+                lambda node, context: (
+                    context.role == "layout"
+                    and TensorOptionalSlotPattern._slot_kind(node, context) == "layout"
+                ),
                 LayoutPattern(),
             ),
             ConditionPattern(
-                "role == storage",
-                lambda node, context: context.role == "storage",
+                "role == storage and value is storage",
```
zhen8838: Why is it judged whether it is Storage in PlacedLayoutPattern?

src/tilefoundry/visitor_registry/access_relation.py
```diff
@@ -784,6 +784,32 @@ def _reached_image(
...
+    """How many points *image* holds, or None when that is not a finite number.
+
+    A box is counted from its bounds: every axis is an independent bounded
+    interval, so the product of their lengths is the answer and ISL never has to
+    enumerate. Enumeration costs one step per point, which a reshaped dimension
+    of a million elements makes unusable, and the reshape and transpose chains a
+    real model is written as reach this with exactly that shape. Anything that is
+    not a box falls back to enumerating.
+    """
+    image = image.coalesce()
```
zhen8838: Didn't you say that isl has a card function?

tests/installed/smoke_analyze.py
```diff
@@ -33,6 +33,28 @@ def main(x: Tensor[(N,), "f32"]):
 '''
 
 
+_DYNAMIC_TRIP_MODULE = '''
```
zhen8838: This dsl should be put in the test fixture and used here.

src/tilefoundry/parser/pattern_nodes.py
zhen8838: Redundant — removed, and it took two dead branches with it.

They were added before the layout branch carried `_states_a_layout_slot`, when a branch could be handed a slot it had to refuse. Now the lookahead settles which arity branch runs, so `context.role` reaching this pattern is already decided; checking the shape again only hid where that decision is made.

With the precise roles no longer re-testing the node, nothing passes `role == "layout_or_storage"` any more, so those two branches are gone as well. `_slot_kind` stays as the lookahead's classifier.

source 927 passed / 4 skipped, installed 168 passed.

src/tilefoundry/visitor_registry/access_relation.py
zhen8838: I said the opposite — this binding does **not** have it, which is why the fast path exists.

Re-checked on isl-python 0.1.11:

```
isl.card               False
isl.pw_qpolynomial     False
isl.set.card           False
[n for n in dir(isl) if "card" in n.lower()]  ->  ["_discard_error"]
```

Upstream ISL in C does have `isl_set_card` (returning an `isl_pw_qpolynomial`), but this binding does not surface it or the polynomial type it returns. So there is no way to ask for a cardinality symbolically here; `count_val` enumerates, one step per point.

Hence `point_count`: read the product off the bounds where set semantics make that exact, and fall back to enumerating otherwise. If the binding grows `card` later, that fallback is the line to replace.

tests/installed/smoke_analyze.py
zhen8838: Moved to `tests/fixtures/diagnostics/dynamic_trip_count.py`, beside `tuple_projection_diagnostic.py`, and reached through a conftest accessor like every other fixture.

You were right that it belonged there — I had kept it inline because `_BAD_MODULE` and `_OPEN_MODULE` in this file are inline, but those are malformed-on-purpose snippets, while this one is a real program that happens to be refused. The diagnostics fixture directory is where that belongs.

Side effect worth noting: the source inventory picks it up automatically, so the suite went 926 -> 927.

src/tilefoundry/visitor_registry/access_relation.py
zhen8838: Correcting myself: my previous reply claimed upstream ISL has `isl_set_card` returning an `isl_pw_qpolynomial`. That is wrong — I did not check, and you did. No such symbol exists in the source tree.

What is actually there:

- `isl_set_count_val` — declared in `include/isl/set.h`, implemented in **`isl_scan.c`**. This is the `count_val` we call.
- `isl_set_count` / `isl_set_count_upto` — also `isl_scan.c`, declared in a private header.
- `isl_set_size` — `isl_map.c`, and it is the size of the internal representation, not a cardinality. Not interchangeable.
- `test/test_isl.py` checks `hasattr(a, "card")`, but the generated Python interface defines no `card`.

The file it lives in makes the point better than my original argument did: counting a set in ISL *is* a scan. There is no symbolic cardinality to reach for, in C or in the binding — so the fallback is not a binding gap that might close, it is the only thing ISL offers. Symbolic counting is barvinok's job, not ISL's, which is where I must have got `card` from.

That strengthens the fast path rather than weakening it, but the reason I gave was wrong. I will put the accurate version in `point_count`'s docstring so nobody goes looking for an API that is not there.

PR #151, opened 2026-09-03
refactor(ir): make the execution domain a first-class MeshScope region

zhen8838:
## Why
- Who-runs-this was derived from `ShardLayout.mesh` and forced to equal result placement; every analysis re-guessed the execution domain from layouts, each with its own bugs.

## What
- `MeshScope` HIR region (mesh + body): scope says who runs, layout says where the result sits; results reach consumers through escaping values (single value, or Tuple + TupleGetItem).
- `MeshScope` is isolated from above: its body references its own params and nothing bound outside; every outside value enters through explicit `args` (RVSDG-style region inputs), so nesting composes structurally and no reach-dependent rule remains.
- Parser: mesh declarations are values, `with` opens a region bound through the lexical frame; a region value made outside a scope cannot be re-bound through it without crossing an operation (ParseError, diagnostics fixtures included).
- Checks: per-operand visibility (covered-by-scope, storage-reaches) with `Reshard` the only exemption; `check_topology` bounds per-level positions; `composed()` replaces same levels / concatenates new ones, scaling strides and offsets, rejecting partial overlap.
- Compute cost: total = per-unit x positions of the enclosing scope, so replicated work is counted.
- Printer round-trips regions byte-for-byte; a region consumed by a sibling scope prints at function level.
- Examples repaired: nemotron declares the machine's real CTA capacity; minicpm3/qwen3.5 import debt fixed.

## Contract
- `ExecutionDomainMetadata` and the "mesh structure derived purely from ShardLayout.mesh" rule are removed; eight spec pages updated. Cost totals now count replicated work (four fixtures' flops changed accordingly).

## Risk
- Lowering MeshScope to TIR is a follow-up plan; escape analysis sees direct bindings only (documented in the plan).
- TIR-side scopes keep their original stack representation; unifying them with HIR regions belongs to the lowering plan.


docs/spec/hir.md
```diff
@@ -412,6 +416,29 @@ node's stamped `.type`.
 Parser-side rules: see
 [parser §3](./parser.md#3-implementation-overview).
 
+### 1.2.1 `MeshScope`
+
+```python
+class MeshScope(Expr):
+    """Represent the execution domain of one structured HIR region."""
+
+    mesh: Mesh
```
zhen8838, line 425: What is described here is different from the final code.

docs/spec/hir.md
```diff
@@ -1417,6 +1444,14 @@ class RoPE(Op):
 `ShardLayout` and `Mesh` are type-system constructs, not Expr inputs
 ([shard §5](./shard.md#5-mesh)).
 
+##### MeshCoord
+
+`MeshCoord(mesh, axis)` is a value-form Op that returns the coordinate of the
```
zhen8838: The MeshCoord description does not write spec according to the format described by op.

docs/spec/parser.md
```diff
@@ -212,6 +212,16 @@ function              ::= 'def' name '(' signature ')' ('->' return-type)? ':' b
 ```
 <!-- parser-grammar:end -->
 
+Mesh declarations are values. `Mesh(...)` constructs the compile-time domain
```
zhen8838: This paragraph should be moved as 1.x section, and why hasn't the syntax been modified? Didn't it regenerate?

docs/spec/shard.md
```diff
@@ -206,6 +206,22 @@ domain, so the composition inherits them from `inner`. Therefore, when an
 outer `ShardLayout` binds a `ComposedLayout`, a `Split(k)` attr still
 references the composition's stable `shape` / `domain_rank` contract.
 
+Mesh composition uses the following rules:
```
zhen8838: It should not be written here, or you can introduce the composed function separately under the Mesh chapter, and then write these.

src/tilefoundry/analysis/compute_cost.py
```diff
@@ -183,6 +222,23 @@ class ComputeCostContext(AnalyzeContext):
...
 
+    def visit(self, expr: Expr, ctx: ComputeCostContext):
+        """Clear inherited scope state when entering a Call's operand graph."""
+        if isinstance(expr, Call):
+            ctx = call_operand_context(ctx)
+        return super().visit(expr, ctx)
+
+    def visit_MeshScope(self, expr: MeshScope, ctx: ComputeCostContext) -> None:
+        """Carry the region's execution multiplicity into each contained Call."""
+        mesh = entered(ctx.inherited, expr.mesh)
```
zhen8838: Why is it enter? Isn't it just composed and then assigned the new mesh to the context?

src/tilefoundry/analysis/compute_cost.py
```diff
@@ -200,9 +256,9 @@ def default_visit_leaf(
         if not isinstance(expr, Call):
             return
         ctx.call_count[0] += 1
-        if ctx.whole is None or ctx.local is None:
-            raise AnalysisError("compute-cost: visitor context is missing cost contexts")
-        record = _call_cost_record(expr, ctx.whole, ctx.local)
+        if ctx.local is None:
+            raise AnalysisError("compute-cost: visitor context is missing its cost context")
+        record = _call_cost_record(expr, ctx.local, ctx.scope_positions)
```
zhen8838: What is ctx.scope_positions?

src/tilefoundry/inspection/python_printer.py
```diff
@@ -714,7 +715,11 @@ def iter_exprs(root: Expr | None, seen: set[int] | None = None) -> Iterator[Expr
     yield root
 
 
-def _collect_meshes(fn: HirFunction, *, include_node_types: bool = False) -> dict[int, Mesh]:
+def _collect_meshes(
```
zhen8838, line 718: Why does printer need to collect mesh first?

src/tilefoundry/ir/tir/sync.py
```diff
@@ -120,7 +120,7 @@ def _(call: "Call", ctx: "VerifyContext") -> None:
         raise VerifyError(
             f"T.sync expects a Mesh argument (m or a slice m[...]), got {type(m).__name__}"
         )
-    scope = ctx.mesh_scope
+    scope = ctx.tir_mesh_scope
```
zhen8838: Why do you need to rename this place to tir_mesh_scope?

src/tilefoundry/ir/types/shard/mesh.py
```diff
@@ -164,54 +165,129 @@ def level_axes(mesh: "Mesh") -> tuple[tuple[int, ...], ...]:
     return tuple(found)
 
 
-def composed(meshes: "tuple[Mesh, ...]") -> "Mesh":
-    """One Mesh naming what a stack of nested single-level Meshes names together.
+def layout_parts(mesh: Mesh) -> tuple[tuple, tuple, int]:
```
zhen8838: The name is wrong, or it should be a nested internal function, not exposed

src/tilefoundry/ir/types/shard/mesh.py
```diff
@@ -164,54 +165,129 @@ def level_axes(mesh: "Mesh") -> tuple[tuple[int, ...], ...]:
...
             )
-        below *= size
-    return Mesh(
-        topologies=tuple(topologies),
-        layout=Layout(shape=tuple(shape), strides=tuple(strides)),
-        names=tuple(names),
-    )
+
+
+def entered(current: Mesh | None, mesh: Mesh) -> Mesh:
```
zhen8838: Don't use this function, at least the name is unreasonable.

src/tilefoundry/ir/types/shard/scope_match.py
```diff
@@ -33,4 +34,35 @@ def mesh_scope_matches_required_scope(current: Mesh, required: Mesh) -> bool:
     return cur_layout.shape == req_layout.shape and cur_layout.strides == req_layout.strides
 
 
-__all__ = ["mesh_scope_matches_required_scope", "states_consistent_positions"]
+def covered_by_scope(mesh: Mesh, current: Mesh) -> bool:
```
zhen8838, line 43: What is the function of this file used for?

tests/ir/test_mesh_coord.py
```diff
@@ -0,0 +1,75 @@
+"""Asking which unit this is: one node, and what every walk makes of it."""
+
+from __future__ import annotations
```
zhen8838: Delete, this should be put in the op test.

tests/ir/test_mesh_scope.py
```diff
@@ -0,0 +1,258 @@
...
+
+    (inner,) = expr_children(outer)
+    assert isinstance(inner, MeshScope) and inner.mesh == THREAD
+    assert leaf in collect_exprs(outer)
+
+    rebuilt = ExprCloner().visit(outer, None)
+    assert rebuilt is outer
+
+
+class _SwapLeaf(ExprCloner):
```
zhen8838: Delete, it should be covered by dsl's parser and analysis test.

PR #158, opened 2026-09-05
feat(runtime): add the warp, mbarrier and TMA op surfaces

zhen8838:
## Why
- A persistent cooperative decode kernel needs three primitive groups this
  runtime does not have: warp-scoped exchange, the Hopper shared-memory barrier
  object, and the bulk asynchronous copy. `async_copy.py` is `cp.async`, not
  TMA; `sync.py` has only `Sync` and `SyncBarrier`; and the only warp shuffle in
  the tree was one `__shfl_xor_sync` private to `ops::reduce`.
- This is the first milestone of replacing an example's generated tilelang mega
  kernel with handwritten CUDA. The ops land in the public runtime rather than
  in the example so the next model does not have to copy them back out.

## What
- `ops::shuffle_xor`, `ops::shuffle_elect`, `ops::warp_reduce`;
  `ops::mbarrier_{init,arrive,arrive_expect_tx,expect_tx,try_wait_parity,
  wait_parity,invalidate}`; `ops::tma_bulk_copy` and `ops::tma_bulk_bytes`.
  One public entry each, implementations under `ops/<op>/`.
- `ops::reduce`'s intra-warp tier folds through `ops::warp_reduce`, so
  `__shfl_xor_sync` is spelled once in the repository instead of twice.
- Matching TIR ops under `ir/tir/cuda/` with types and verifier rules.
- `tests/integration/test_runtime_device_ops.py` drives every entry from a
  hand-written kernel and compares to torch, including a three-deep staged ring
  that exercises the phase parity in both states.

## Contract
- `docs/spec/runtime.md` gains §3.7–3.9 and `docs/spec/tir.md` gains the three
  op groups.
- **None of the three admits a tier.** A warp shuffle's operand is a register
  scalar; an mbarrier is a shared-memory word, not a sharded tensor; and the
  hardware's two bulk forms differ in their *operands* — a size against a
  host-encoded `TensorMap` — rather than in a tier a trait could select. Each is
  a single implementation behind a single entry, and the caller passes no
  selection. §3's one-entry rule is satisfied; it does not require an op to have
  more than one implementation.
- **The TIR ops carry no CUDA emit.** The compile side is being rebuilt; an emit
  written now would be rewritten by that work while these definitions would not.
  Each spec entry says so and names the `ops::` entry it corresponds to.

## Risk
- Because there is no emit, the established way to check a runtime op against
  torch (`test_mma_tir_handwritten.py`: author TIR, compile, let codegen emit
  the call) is not open to these. The new test pins the *implementation*; what
  ties it to the TIR definition is review, not the machine. Wiring the emit is
  what would pin the pair, and that waits on the TIR backend rebuild.
- `ops::tma_bulk_copy` implements only the rank-1 form. The tensor forms
  (`cp.async.bulk.tensor.Nd`) are absent because nothing calls them.
- The wheel ships neither `include/` nor the vendored CuTe, so a handwritten
  kernel outside this repository cannot yet reach these headers. `examples/` is
  not distributed either, so nothing in tree depends on that today.

### DIFF
--- a/CLAUDE.local.md
+++ b/CLAUDE.local.md
@@ -1,0 +1,1 @@
+/home/qihang.zheng/TileFoundry/CLAUDE.local.md
--- a/docs/spec/runtime.md
+++ b/docs/spec/runtime.md
@@ -1036,3 +1036,142 @@ __device__ void copy_async(TSrc const& src, TDst& dst);
     `cp_async_commit` and `cp_async_wait`.
   - Runtime implementation details such as vector width, tail handling, and
     architecture fallback live in code comments, not this spec entry.
+
+### 3.7 `tilefoundry::ops::warp_*` (warp-scoped exchange)
+
+Three ops, one public entry each, and **no tiers**: their operand is a
+register-resident scalar, which carries no `ShardLayout`, so there is nothing
+for a compile-time tier selection to read. The one-entry rule of
+[§3](#3-runtime-ops) requires one entry per op; it does not require every op to
+have more than one implementation.
+
+`warp.cuh` is included before `reduce.cuh` so that `ops::reduce`'s intra-warp
+tier folds through `warp_reduce` rather than spelling the butterfly a second
+time.
+
+```cpp
+/**
+ * @brief Exchange a value with the lane whose id differs by lane_mask.
+ * @param value this lane's contribution
+ * @param lane_mask the lane-id bits exchanged across
+ * @param member_mask the participating lanes
+ */
+template <class T>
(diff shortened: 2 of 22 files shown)


zhen8838:
Closing: opened after the first milestone only. The plan replaces a 1564-line generated tilelang mega kernel with handwritten CUDA across five milestones, and only the runtime op surfaces are done. Reopening when the whole thing lands, so this is reviewed once against the finished work rather than as a library with no user yet. The branch keeps the commit.

PR #159, opened 2026-09-05
feat(examples): replace the nemotron tilelang mega kernel with handwritten CUDA

zhen8838:
## Why
- `examples/nemotron_3_5_lightning_30b_a3b-tilelang` runs its decode step as one
  launch of 1564 generated lines of TileLang. This replaces that with handwritten
  CUDA on the path `examples/granite_4_0_h_small-cuda` already proved: a `.cu`,
  compiled on first use, launched from `@runtime_func`.
- Three primitive groups the runtime did not have had to come first: warp-scoped
  exchange, the Hopper mbarrier, and the bulk asynchronous copy.

## What
- **Runtime.** `ops::{shuffle_xor, shuffle_elect, warp_reduce}`,
  `ops::mbarrier_*`, `ops::tma_bulk_copy`. One public entry each under
  `include/tilefoundry/runtime/cuda/ops/`, implementations under `<op>/`, tested
  against torch from a hand-written kernel. `ops::reduce`'s intra-warp tier now
  folds through `warp_reduce`, so `__shfl_xor_sync` is written once.
- **TIR.** The matching ops under `ir/tir/cuda/` with types and verifier rules.
- **Kernel.** `kernels/nemotron.cu`: all 52 layers, the closing norm and the head
  in one cooperative launch of 132 CTAs (`NEMO_IMPL=cuda`). Every stage is a
  `__device__` function that a launch-per-stage path (`cuda-stages`) calls too,
  so the two cannot drift.
- **Harness.** `check_all.py` could not run against the current CLI at all;
  `dump_acts.py`, `check_layer.py`, `check_moe.py` and `compare_impls.py` are
  new. The weight layout moves to `packing.py` so the CUDA paths no longer need
  TileLang installed.

## Contract
- `docs/spec/runtime.md` §3.7–3.9 and `docs/spec/tir.md` gain the three groups.
  None admits a tier — a warp shuffle's operand is a register scalar, an mbarrier
  is not a sharded tensor, and the two bulk forms differ in operands, not tiers.
- The TIR ops carry **no codegen emit**: that chain is being rebuilt.
- `NEMO_IMPL` defaults to `mega` still. The CUDA path is opt-in.

## Risk
- **It is 2.19x slower** (153.6 tok/s against 335.9 at ctx 32, same session).
  `reports/WHY_SLOWER.md` takes it apart: which layers, what the grid barrier
  costs, what is fixable, and the four hypotheses that measurement killed.
- No `ncu` on this machine (ERR_NVGPUCTRPERM), so that report rests on `nsys`
  tracing and wall-clock against byte counts — no occupancy, no stall reasons.
- Nothing in CI exercises `examples/`, and the ruff config excludes it. The
  evidence for the kernel is the reports in that directory, not the test suites.
- Because there is no emit, what ties each TIR op to the implementation it names
  is review, not the machine.

Verified: `tilefoundry-verify source` 983/0 and `installed` 161/0.

### DIFF
--- a/CLAUDE.local.md
+++ b/CLAUDE.local.md
@@ -1,0 +1,1 @@
+/home/qihang.zheng/TileFoundry/CLAUDE.local.md
--- a/docs/spec/runtime.md
+++ b/docs/spec/runtime.md
@@ -1036,3 +1036,142 @@ __device__ void copy_async(TSrc const& src, TDst& dst);
     `cp_async_commit` and `cp_async_wait`.
   - Runtime implementation details such as vector width, tail handling, and
     architecture fallback live in code comments, not this spec entry.
+
+### 3.7 `tilefoundry::ops::warp_*` (warp-scoped exchange)
+
+Three ops, one public entry each, and **no tiers**: their operand is a
+register-resident scalar, which carries no `ShardLayout`, so there is nothing
+for a compile-time tier selection to read. The one-entry rule of
+[§3](#3-runtime-ops) requires one entry per op; it does not require every op to
+have more than one implementation.
+
+`warp.cuh` is included before `reduce.cuh` so that `ops::reduce`'s intra-warp
+tier folds through `warp_reduce` rather than spelling the butterfly a second
+time.
+
+```cpp
+/**
+ * @brief Exchange a value with the lane whose id differs by lane_mask.
+ * @param value this lane's contribution
+ * @param lane_mask the lane-id bits exchanged across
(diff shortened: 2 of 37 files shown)


zhen8838:
Closing: the goal is a handwritten kernel no slower than the TileLang one, and this is 2.19x slower (153.6 tok/s against 335.9 at ctx 32, same card, same session). Milestone checkboxes are not the goal. Reopening when the number is met. The branch keeps the work.

PR #160, opened 2026-09-05
feat(runtime): let the shard layouts decide the copy, the contraction and the mma

zhen8838:
## Why

The nemotron example's handwritten decode kernel was CUDA, not TileFoundry: it
hand-rolled every thread mapping, reinterpret-cast rows to `int4`, and carried
eight partial sums per warp. None of that is a thing a kernel written against
this runtime should have to say — the layouts already know the transfer width,
the contraction mesh and the tiling.

Three runtime ops could not express what the kernel needed, so it did not call
them: `tma_copy` took a byte count and a raw pointer, `mma` was one instruction,
and there was no entry for the fused multiply-contract that *is* a matrix-vector
row.

## What

Runtime — the tier, the width and the shape now come off the operand
`ShardLayout`s:

- `ops::tma_copy` takes tensors. A shard layout that coalesces to one static
  contiguous run takes `cp.async.bulk`; anything else takes the element path.
  Both complete on the barrier, so the caller never learns which ran.
- `ops::copy` is a new entry. The move width is `max_common_vector` of the two
  *projected* layouts, capped at 16 bytes; `copy_vector_elems` reports it. This
  replaces a hard-coded 128-bit path guarded by an O(N) address scan.
- `ops::mma` gains a tile tier: rank-2 static operands loop the atom, a lane's
  gathered fragments take the single instruction. There is no transpose flag —
  an `(N, K)` view of a `(K, N)` buffer is a stride.
- `ops::dot` is new: `sum(lhs * rhs)`, warp- or block-scoped from the mesh.
- `warp_reduce` gains a `Width`, for a per-row fold over a tile laid out several
  rows to a warp.
- `detail::local_layout` projects a `ShardLayout` through its mesh in the type
  system, which is what makes any of the above a compile-time answer.

Example — `kernels/nemotron.cu` is rewritten against those entries, with every
layout in `kernels/shards.cuh`. No thread index appears in arithmetic. The
attention scan is now the shape the TileLang kernel uses: one `ops::mma` for a
`(16, 128)` score block, an online softmax, a second `ops::mma`.

## Contract

- `docs/spec/runtime.md` §2.8 and §§3.6–3.13 restate the copy family, `dot`,
  `mma` and the retitled `tma_copy`. `ops::tma_bulk_copy` / `tma_bulk_bytes` are
  gone; the TIR op `TmaBulkCopy` is now `TmaCopy` and no longer rejects a
  transfer off the 16-byte grain — the grain belongs to one instruction, and the
  definition layer may not carry a tier.
- `warp_reduce`'s new `Width` defaults to 32; every existing call is unchanged.

## Risk

- 4.32 ms a token against the TileLang kernel's 2.977, so **1.45x slower** and
  still not the default (it was 2.19x). `reports/WHY_SLOWER.md` carries the
  attribution, the four changes that closed the gap, and five things that were
  tried and measured worse — including an arena refactor that frees 78 KB and
  fits two CTAs an SM and is 23% slower for reasons three experiments failed to
  isolate.
- The rewrite surfaced a pre-existing double-count in the routed down
  projection: per-CTA row counts rounded up to whole tiles overlapped, and that
  stage accumulates atomically. Split by `tile_span` instead.


zhen8838:
Closing in favour of a focused replacement: this branch grew to 46 commits spanning the parser, the analysis passes, the tutorial and the CUDA runtime, which is not reviewable as one change. The runtime ops work is being reopened on its own branch as a single commit covering the runtime, its TIR nodes, its tests and its spec, with no example changes.

PR #161, opened 2026-09-08
feat(runtime): close ops on the seven entries the layouts describe

zhen8838:
## Why

- `ops::` had grown entries that the layout system cannot describe, and paths
  that answered a wrong result where they should have refused. Ten pointwise
  entries encoded broadcast patterns and extents in function names; the
  `mbarrier_*` wrappers, the tier predicates and the `mma_acc_*` helpers took
  pointers, ints and types rather than tensors. Meanwhile `ops::dot` and
  `ops::tma_copy` had no TIR node at all, so no compiler output could reach
  them and nothing guarded them.

## What

- **One pointwise entry.** `fill`, `unary`, `binary`, the four `binary_bcast_*`
  variants, `cast`, `clamp` and `rmsnorm` become `elementwise(dst, fn, src...)`.
  A broadcast is a stride-0 mode on the operand's layout; the count is
  `size(local(dst))`. Eight impls become one.
- **Non-ops leave `ops/`.** The seven `mbarrier_*` wrappers, `mma_is_tile`,
  `tma_copy_is_bulk` and the `mma_acc_*` helpers are gone; the mbarrier
  instructions are written at their call sites.
- **Every op is reachable.** New TIR nodes and emitters for `Dot`, `TmaCopy`
  and `MBarrier*` (the last lowering to its PTX instruction, not to a runtime
  entry).
- **Wrong usage stops compiling.** Nine fail-open paths become
  `static_assert`s and the `if constexpr` chains are exhaustive: a workspaceless
  cross-warp `reduce`, a `dot` on a sub-warp mesh, `mma` operands that are
  neither tile nor atom, an attrs tuple whose length misses the mesh rank, and
  more.
- **Tests.** One module per op covering each tier on GPU, plus a
  compile-failure harness of 25 cases pinning each new constraint.

## Contract

- `ops::` is closed on seven entries (runtime §3): `elementwise`, `copy`,
  `copy_async`, `reduce`, `dot`, `mma`, `sync`, `tma_copy`. The warp primitives
  move to their own chapter (§4) as not-ops.
- `ops::tma_copy` no longer selects a tier statically. The static strided path
  was reachable and every layout that selected it computed a wrong answer — one
  case read 511 of 512 elements out of bounds. It is replaced by five asserted
  preconditions; `Strided` remains as the run-time hand-off off the 16-byte
  grain.
- A register `ShardLayout` now carries storage strides, so `local()` has no
  storage-specific branch and shard §7.1.2's rule holds as written.
- `ReduceKind` gains `MIN`, matching the runtime's five `reduce_traits`
  specialisations; `MAX` previously raised `KeyError` at emission.
- `Partial` renders its reduction into `shard::P<Op>` instead of `P<void>`.

## Risk

- A codegen RMEM allocation now uses a named array and a pointer engine. The
  previous `make_tensor<T>(layout)` gave `ShardTensor` an engine it held by
  value, so writes through `local()` landed in a copy; a test written to record
  that defect now passes.
- `docs/SPEC-RULES.md`'s numbering rule is relaxed to permit renumbering behind
  the refs lint, which this change needed twice.
- Not covered: `Partial.reduction` is still a free string in the IR and is only
  rejected at codegen, and `ir/hir/math/binary.py` refuses a Partial operand to
  elementwise min/max, which is conservative rather than wrong.


docs/spec/runtime.md
```diff
@@ -628,6 +628,18 @@ under `tilefoundry/runtime/cuda/runtime.cuh` (the CPU surface under
 (`runtime/<target>/…`), no intermediate `target/` segment. Generated code MUST
 include only the umbrella header and MUST NOT include target subheaders directly.
 
+`runtime.cuh` also re-exports the CuTe primitives generated code is written in
```
zhen8838, line 631: There is no need to write this.

docs/spec/runtime.md
```diff
@@ -636,12 +648,18 @@ include only the umbrella header and MUST NOT include target subheaders directly
  */
 enum class TopologyScope {
   cta,          ///< maps to blockIdx
+  warp,         ///< a warp of the block; named here, not queryable (below)
```
zhen8838, line 651: Warp scope is deliberately not added.

docs/spec/runtime.md
```diff
@@ -680,43 +698,133 @@ always a runtime query returning the current execution instance id.
...
+  static constexpr TopologyScope scope = Scope;  ///< the level, and no extent beside it
+};
+
+/**
+ * @brief A device mesh: one topology level bound to a CuTe mesh layout.
+ * @tparam TTopo a `Topology<Scope>`
+ * @tparam TMeshLayout a `cute::Layout`, or a `cute::ComposedLayout` for a slice
  */
-template <class MeshLayout, TopologyScope... Topos>
+template <class TTopo, class TMeshLayout>
```
zhen8838, line 714: Why do we need to change the mesh structure? The original is reasonable.

docs/spec/runtime.md
```diff
@@ -804,10 +912,43 @@ void copy(ShardTensor<T, GL, SL> const& src, DT& dst);
...
+template <class TS, class GLS, class SLS, class TD, class GLD, class SLD>
+void copy(ShardTensor<TS, GLS, SLS> const& src, ShardTensor<TD, GLD, SLD>& dst);
+
+/**
+ * @brief Copy the full tensor, plain → plain.
+ * @param src the source tensor
+ * @param dst the destination tensor
+ */
+template <class ST, class DT>
+void copy(ST const& src, DT& dst);
```
zhen8838, line 930: Why is copy written here and copy is written below?

include/tilefoundry/runtime/cuda/ops/elementwise/tags.h
```diff
@@ -0,0 +1,112 @@
...
+/// Tags and nothing else: values a caller hands an op as its callable, so they
+/// are not ops and do not sit beside the entries in ``ops/``. ``reduce`` and
+/// ``dot`` specialise ``reduce_traits`` on some of the arity-2 tags, which is
+/// what makes the elementwise maximum of two tensors and the maximum over an
+/// axis one name rather than two that must be kept meaning the same thing.
+#pragma once
+
+/// Arity 1.
+
+struct rsqrt_op {
```
zhen8838, line 13: Why do you add tags here? It should be to add primitive ops in runtime ops, which are used as primitive.

tests/ir/tir/test_elementwise_broadcast.py
```diff
@@ -0,0 +1,146 @@
...
+    @prim_func(target=_CUDA)
+    def col_bcast_device(
+        a: Tensor[(32,), "f32"], r: Tensor[(4,), "f32"], out: Tensor[(32,), "f32"]
+    ):
+        with Mesh((Topology("thread", 32),), Layout(shape=(32,), strides=(1,)), ("t",)) as m:
+            a_view = T.tensor_view(a, layout=_bc((32,), (1,), m))
+            r_view = T.tensor_view(r, layout=_bc((4,), (1,), m))
+            out_view = T.tensor_view(out, layout=_bc((32,), (1,), m))
+            lhs = T.alloc_tensor(
+                TensorType(
```
zhen8838, line 68: Why don't you write grammar sugar?

tests/ir/tir/test_mma_tile_tier.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+def test_the_tile_tier_reaches_the_same_entry_as_the_atom() -> None:
+    """No tier and no atom name on the line -- only the three operands.
+
+    Which of the two tiers runs is the operands' layouts to say, so the emitted
+    call for a broadcast tile is character-for-character the shape the gathered
+    fragments emit.
+    """
+    from tilefoundry.codegen.cuda.module import emit_cuda_module  # noqa: PLC0415
```
zhen8838, line 123: Why import internally?

tests/ir/tir/test_sync.py
```diff
@@ -21,10 +24,11 @@
 
 
 def test_sync_barrier_forms_emit_expected_cuda() -> None:
```
zhen8838, line 26: Why do you need to test the cuda emitted? Write dsl directly and test compile execution

PR #162, opened 2026-09-09
refactor(runtime): close the gaps PR #161's review and discussion turned up

zhen8838:
## Why
- PR #161 closed `ops::` on seven entries. Its review left eight items, and the discussion that followed turned up one more defect (`ops::rmsnorm` split into the emitter) plus three calls the original judgement got wrong.

## What
- Runtime spec follows hir §1.3's shape: a numbered group per source directory, unnumbered entries under it. Deleted the CuTe re-export section and §2.8 `tilefoundry::copy`, both describing interfaces that no longer exist.
- `ops::rmsnorm` is an op again: one dependency chain, one scalar accumulator per row, `M`/`K` off the destination's shard layout instead of the emitter.
- Op tags move to `ops/primitive/{unary,binary}.h`; `mean_op` / `absmax_op` stay beside `reduce` because they carry no `operator()`.
- A host entry may launch several device functions in order, and codegen splits translation units by declared topology domain, so variants keep their own block sizes in one build.
- Tests are organised by dialect x target; `tests/ir/tir/` is gone.
- C++ `Mesh` is variadic over topology levels, matching the IR; warp geometry is answered once by `MeshWarpView`; `sync` and `reduce` state their tier through a `plan`.

## Contract
- **Breaking.** `dot(lhs, rhs, dst)` without a workspace no longer compiles on a mesh wider than one warp. It used to return one warp's partial sum as the answer, the same defect `reduce` already fixed.
- `T.tensor_view` accepts an allocated `ShardTensor` as `memory`, reprojecting the same storage through its engine.
- A `Mesh` naming several topology levels is rejected at compile time; the axis assignment rule is not implemented and has no caller today.
- `ops::rmsnorm` returns to the runtime surface; `ops::` now holds nine entries.

## Risk
- Multi-level mesh lowering is refused, not supported. The first real caller adds the rule and its test together.
- `Module.topologies` and a `launch`'s block are not cross-checked; device grid/block comes from the body mesh while the host launches what it was told.
- `reduce_traits` still routes every accumulation through f32.


include/tilefoundry/runtime/cuda/layout/shard_layout.cuh
```diff
@@ -35,3 +161,188 @@ template <class Reduction> struct P {
...
+///
+/// The stride order is what turns a linear instance id into a coordinate:
+/// ``(8, 32)`` row-major gives ``(id / 32, id % 32)``, so a thread mesh names
+/// its warps first and its lanes last.
+///
+/// A mesh wider than its level is refused here rather than read out of range
+/// later: ``program_shape<Scope>()`` is what the launch states the level is, so
+/// nothing else has to be told, and every ``shard_mesh_coord`` on an oversized
+/// mesh indexes past the last instance the level has.
+template <TopologyScope Scope, class Extents>
```
zhen8838: Does the lint of our Comment governance have no effect on this?

include/tilefoundry/runtime/cuda/ops/copy/copy_impl.h
```diff
@@ -4,26 +4,93 @@
...
-        if (C > 1) {
-            for (; i + C <= n_src; i += C)
-                __pipeline_memcpy_async(&d(off + i), &s(i), C * ES);
-        }
+        using value_type =
+            typename cute::remove_cvref_t<decltype(d)>::value_type;
+        constexpr int bytes = detail::async_bytes<decltype(s), decltype(d)>();
+        static_assert(
+            bytes > int(sizeof(value_type)),
+            "ops::copy_async: these two views share no run wide enough for "
```
zhen8838: Is there any way to limit the string size of assert?

include/tilefoundry/runtime/cuda/ops/mma/mma_impl.h
```diff
@@ -14,14 +14,32 @@ template <class T> __device__ uint16_t as_u16(T const &x) {
     return out;
 }
 
+/// Geometry of the ``m16n8k16`` atom, in the order its PTX operands take.
+///
+/// Every index the tile tier computes comes from here rather than from a
+/// literal at the use site, so the four places that agree on this map -- the A
+/// gather, the B gather, the accumulator read and its write-back -- cannot
+/// drift apart.
+struct Sm80_16x8x16 {
```
zhen8838: Why do you need this? Cute has a complete set of mma atom structures.

include/tilefoundry/runtime/cuda/ops/sync/sync_impl.h
```diff
@@ -23,26 +23,209 @@ __device__ __forceinline__ void grid_barrier(unsigned int *bar) {
...
+    __device__ void operator()(unsigned int *bar) const {
+        if (bar != nullptr)
+            grid_barrier(bar);
+        else
+            cooperative_groups::this_grid().sync();
+    }
+};
+
+/// Every thread of the block.
+struct Block {
```
zhen8838, line 140: This does not support half of the block and half of the warp for synchronization.

include/tilefoundry/runtime/cuda/runtime.cuh
```diff
@@ -15,71 +15,103 @@
...
+template <class...> inline constexpr bool dependent_false_v = false;
+
+/// The fixed enumeration of [runtime
+/// §2.1](docs/spec/runtime.md#21-topologyscope). Two of the four are levels
+/// this backend can project; ``warp`` and the ``scope_count`` sentinel are not,
+/// which is why a dispatch on scope is exhaustive rather than trailing an
+/// ``else`` that stands for ``thread``.
 enum class TopologyScope {
     cta,
     warp,
```
zhen8838: I said that there is no warp in topology scope.

scripts/comment_hygiene_lint.py
```diff
@@ -253,6 +253,27 @@ def _c_comments(text: str) -> list[tuple[int, str, str, bool]]:
     return found
 
 
+def _next_code_line(text: str, after: int) -> str:
```
zhen8838, line 257: Why are there so many comments on the cpp section now?

src/tilefoundry/codegen/cuda/tir/nn/mma.py
```diff
@@ -11,7 +13,7 @@
 from tilefoundry.ir.tir.cuda.nn.mma import Mma
 
 _MMA_RUNTIME = {
-    "SM80_16x8x16_F32BF16BF16F32_TN": "tilefoundry::ops::mma_sm80_16x8x16_bf16",
+    "SM80_16x8x16_F32BF16BF16F32_TN": "tilefoundry::ops::mma",
```
zhen8838: Why not mma_sm80_16x8x16_bf16?

src/tilefoundry/codegen/cuda/tir/stmts/mesh_scope.py
```diff
@@ -12,40 +12,109 @@
...
     check at lowering entry.
     """
-    validate_cuda_topology_levels(target, (t.name for t in mesh.topologies))
-
+    validate_cuda_topology_levels(
+        target, (_resolved(t).name for t in mesh.topologies)
+    )
 
 
+def mesh_geometry(mesh: Mesh) -> tuple[tuple, tuple, int]:
```
zhen8838: Why must mesh_geometry?

include/tilefoundry/runtime/cuda/layout/shard_layout.cuh
zhen8838: Fixed in 496808b: every comment this change adds is one line, and `shard_layout.cuh` is 216 -> 113 lines with the mesh half moved to `layout/mesh.cuh`.

include/tilefoundry/runtime/cuda/ops/mma/mma_impl.h
zhen8838: Fixed in 496808b: the atom geometry is read off `cute::MMA_Traits<SM80_16x8x16_F32BF16BF16F32_TN>` -- `Shape_MNK` for the tile and `CLayout` for the lane/value map -- instead of being restated.

include/tilefoundry/runtime/cuda/runtime.cuh
zhen8838: Fixed in 496808b: `TopologyScope` is `cta`, `thread`, `scope_count`. A warp is an axis of a thread mesh's layout, and `warp_tiling` reads it off the ids.

scripts/comment_hygiene_lint.py
zhen8838: Fixed: 2074 added C++ lines now carry 265 comment lines, one per site. b880fe8 did the same for this change's Python docstrings, 671 lines down to their summaries.

src/tilefoundry/codegen/cuda/tir/stmts/mesh_scope.py
zhen8838: Fixed: `mesh_geometry` is gone. A mesh scope emits the mesh type and one object, and the views under it name that object rather than rebuilding an equal mesh each.

include/tilefoundry/runtime/cuda/ops/sync/sync_impl.h
zhen8838: Supported now, and checked: a contiguous run of 2, 3 or 4 warps classifies as Tier::named and syncs on bar.sync, whether or not it starts at warp 0. `tests/ops/tir/test_sync.py::two_warps` is `full[2:4]` of a 128-thread block. What is refused is the other half of the sentence -- a mesh whose lanes run less than a warp and repeat across warps, which reaches neither bar.sync, that counts whole warps, nor one __syncwarp. It says so by name rather than guessing.

src/tilefoundry/codegen/cuda/tir/nn/mma.py
zhen8838: The op is `Mma` and `SM80_16x8x16_F32BF16BF16F32_TN` is the atom, which is CuTe's own name for it: `mma_sm80_16x8x16_bf16` would drop the F32 accumulator and the TN layout, and this spelling is the one `cute::MMA_Traits<>` is instantiated with, so the two cannot drift.

include/tilefoundry/runtime/cuda/ops/copy/copy_impl.h
zhen8838: Added in 183b15d. clang-format cannot do it -- `BreakStringLiterals` folds a long message to fit the 80-column limit and takes nothing off its length -- so the check went into comment_hygiene_lint, next to the rule that already finds a static_assert to say prose belongs in its message. Adjacent literals are one message in C++, so it measures their concatenation. The limit is 100 characters: 76% of the 69 messages already sat under 80, and the seven over it each carried a clause the assertion's own name already said.

PR #165, opened 2026-09-10
feat(runtime): load each program its own shard of a checkpoint

zhen8838:
## Why
- A host runtime reading a full checkpoint had no way to say which program it was, so every card read every weight whole.
- The native side refused a mesh naming several topology levels, and refused a tensor axis two mesh axes cut, so the layouts the Python side accepts could not be lowered.
- Three rounds of review, twenty-six findings. Most were one complaint from different angles: a name that does not say what it asks, or something stated twice.

## What
- `Placement` says which program a host invocation is: a `{level: id}` mapping, fixed when a module is loaded. `ShardTensor.to_local` narrows a canonical weight to that program's part.
- One mesh may name several levels, and several mesh axes may cut one tensor axis. `mesh_coords` answers index-to-coord once per level against that level's own nest.
- A `gpu` topology level, `CudaTarget(device_count=)`, and `ProgramMetaData` passed by value so `program_id<gpu>()` reads what the host supplied.
- Which part of a tensor is one instance's has one implementation per side under one name, `local_layout_and_offset`: host and device read the one layout and the one offset.
- Codegen settles every function's signature before any of them is written. A call site asks the table what it is calling, so a function can be called before it has been emitted, and one signature writes both the forward declaration and the definition.
- A signature is a list of one class per kind of parameter, and what a kind becomes in C++ is a handler registered against `(target, role, type)` — the target being the *callee's*, so a host entry emitting a call into a device function asks CUDA how such a call is written.
- A translation unit is assembled from its functions instead of rendered a second time beside them.

## Contract
- `Placement` is data, not a callable: one loaded module speaks for one program. Serving several cards is several modules.
- `ComputeCostMetadata` and `TrafficMetadata` share one shape and name their levels once in `topologies`. A report states totals under `totals`.
- Two hierarchies no longer share the word "level": execution is `topology_level`, memory is `memory_level`, and the per-TU macro is `TILEFOUNDRY_PROGRAM_TOPOLOGY`.
- A dynamic extent is stated once, as a `DimVar` in the tensor's type. The scalar parameter that carried the same fact under a parsed-back name is gone, and a user's own `foo_shape_2` is no longer mistaken for one.
- The pipeline over a TIR module is one pass. `HirToTirPass`, `BufferizePass` and `tilefoundry.lower` are removed; HIR to TIR becomes its own thing.
- Specialization dispatch happens on the device. The variants share one geometry because a translation unit specializes `program_dim<>` once.
- `runtime.md` is organised by file, one heading per class, and every signature block is read out of the source by `scripts/runtime_spec_surface.py` under a hook that reports a spec that has fallen behind.

## Risk
- Nothing executes cross-card movement. A `Reshard` over a `gpu` axis is priced, not performed.
- Two cards is what CI and local verification cover; nothing here has run on more.
- `Spread` reports a whole that moves with the level asked about while its per-unit shares do not. Pre-existing, now pinned by value so a fix shows up as a deliberate change rather than drift.
- `CudaTarget` still parses and cross-checks inside a hand-written `__init__` on a frozen dataclass. Left alone on purpose; it is a question about how every Target is constructed, not about codegen.


docs/spec/analysis.md
```diff
@@ -170,23 +170,26 @@ class ComputeCostMetadata(IRMetadata):
...
-        flops_per_unit: attribute; Flop count performed by one unit of the analysed topology level.
+        by_unit: attribute; What one unit does, at every declared topology level.
         service: attribute; Result count per service kind, sorted by kind.
-        service_per_unit: attribute; Result count per service kind for one unit of the analysed topology level.
+        unit: attribute; The level this analysis was asked about, out of the ones by_unit holds.
     """
 
     flops: tuple[tuple[str, int], ...] = ()
-    flops_per_unit: tuple[tuple[str, int], ...] = ()
+    by_unit: tuple[tuple[str, UnitWork], ...] = ()
```
zhen8838: What is unitwork? Why don't you be equal to flops?

docs/spec/analysis.md
```diff
@@ -374,7 +377,28 @@ of this analysis.
     domain that fits at once MUST be settled without searching, and one whose
     simultaneously live bytes exceed the capacity MUST be refused without
     searching.
-  - A Call's `whole`, `per_unit` and `operands` MUST state one occurrence. Only
+  - `*_per_unit` and the per-unit share of a traffic level MUST be stated for
```
zhen8838: Please simplify the explanation.

docs/spec/runtime.md
```diff
@@ -612,472 +626,359 @@ def check(candidate: Callable, reference: Callable | None, inputs: tuple, *,
...
 
-## 2. C++ Runtime Surface
-
-The sections follow `include/tilefoundry/runtime/`, one per header. Each
-declaration block is read out of that header by
-`scripts/runtime_spec_surface.py` and written into its `<!-- generated -->`
-region, so a signature is stated once, where it is compiled. What follows a
-region is written by hand: the terms the declarations are spelled in, and the
-decisions two reasonable implementations would differ on.
+### 1.7 `ShardTensor`
```
zhen8838: This class should not be written here. The structure of the runtime spec needs to be re-combed.

src/tilefoundry/analysis/api.py
```diff
@@ -102,6 +103,25 @@ def visit(selector: str, root: str) -> None:
     return tuple(ordered)
 
 
+def _coarsest_measured_level(target, topologies) -> "str | None":
```
zhen8838: Why try catch?

src/tilefoundry/runtime/decorator.py
```diff
@@ -7,16 +7,31 @@
...
+from tilefoundry.ir.types.tensor_type import TensorType
 from tilefoundry.runtime.function import RuntimeFunction
 from tilefoundry.runtime.module import RuntimeModule
 from tilefoundry.runtime.resource import RuntimeResource
+from tilefoundry.runtime.tensor import ShardTensor
 
 _RUNTIME_FUNC_MARK = "_tilefoundry_runtime_func"
 
 
+def _current_device() -> "int | None":
```
zhen8838: Why did you add this function here?

src/tilefoundry/runtime/decorator.py
```diff
@@ -85,16 +104,60 @@ def module(self) -> Module:
...
+        self, resource: RuntimeResource, *, placement: "Placement | None" = None
+    ) -> None:
         self._resource = resource
+        self._placement = placement
         self._bound.clear()
         for child in self.modules:
-            child.load(resource.subtree(child.name))
+            child.load(resource.subtree(child.name), placement=placement)
+
+    def _program(self) -> tuple:
```
zhen8838: Need to understand warp unwarp program

src/tilefoundry/runtime/tensor.py
```diff
@@ -0,0 +1,112 @@
...
+                extent = mesh_shape[mesh_axis]
+                position = next(coord) if extent != 1 else 0
+                tensor_axis = tensor_axes[mesh_axis]
+                if tensor_axis is None:
+                    continue
+                out = _one_of(out, tensor_axis, extent, position)
+        self._check(out, ids, at)
+        return out
+
+    def _check(self, out: torch.Tensor, ids: tuple, at: dict) -> None:
```
zhen8838: What is the function of one of and check?

src/tilefoundry/target/cuda/facts.py
```diff
@@ -84,6 +85,24 @@ def memory_hierarchy(target: CudaTarget, query: object = None) -> MemoryHierarch
     )
 
 
+def _units_per_device(target: CudaTarget, unit: str) -> int:
```
zhen8838: Why is gpu 1 by default?

tests/fixtures/placed/tp_all_to_all.py
```diff
@@ -0,0 +1,40 @@
...
+from tilefoundry import func, module
+from tilefoundry.dsl import Mesh, Tensor, Topology, tf
+from tilefoundry.ir.types.shard import Layout, Split, canonical_shard_layout
+from tilefoundry.ir.types.shard import Mesh as ShardMesh
+from tilefoundry.target import CudaTarget
+
+GPUS, CTAS, R, C = 2, 4, 8, 8
+GPU, CTA = Topology("gpu", GPUS), Topology("cta", CTAS)
+
+_MESH = ShardMesh(topologies=(GPU,), layout=Layout((GPUS,), (1,)), names=("g",))
```
zhen8838, line 22: What is this test used for?

tests/fixtures/placed/tiny_tp_decoder.py
```diff
@@ -0,0 +1,97 @@
...
+@runtime_module(DecoderLayer)
+class DecoderLayerTwin:
+    """The twin is the assertion.
+
+    A body is where the runtime hands one program its data, so the judgement
+    that it is this card's data and no other belongs right here rather than
+    back out in a test that would have to rebuild the expectation.
+    """
+
+    @runtime_func
```
zhen8838, line 80: You haven't checked the data values obtained by different devices.

docs/spec/analysis.md
zhen8838: You are right that the two records did not line up. `flops` was the whole program and `UnitWork` was one unit, in a shape traffic did not share, so the same question was written two ways.

Both are now one shape: `Breakdown[V]` keyed by kind (dtype / service kind / storage level / crossed boundary), each kind a `Spread[V]` of a total and one share per topology level, with the level names written once on the record as `topologies`. `UnitWork`, the `""` key that stood for the whole, the `unit` note, and eight accessor methods are gone.

docs/spec/analysis.md
zhen8838: Compressed to one rule; the reason it exists moved to the `Spread` Terms entry, which is what that type is for. The last sentence about asking for one level or all of them is deleted -- a `Spread` now always carries both, so there is no such question.

docs/spec/runtime.md
zhen8838: Agreed, and it was two problems. §1 was organised by concept while §2 was by file, so the host `ShardTensor` and the device `ShardTensor` stood as peers under one name. §3 and §4 organised the op surface a second time, by op, so a signature and its semantics sat 800 lines apart and §3 rebuilt §2.6 directory structure to say so.

§1 is now by file (`runtime/module.py`, `function.py`, `decorator.py`, ... and the host one is §1.10 `runtime/tensor.py`). §3 and §4 are deleted, their per-op prose merged into each file section and their preambles into §2.6 and §2.7.

src/tilefoundry/analysis/api.py
zhen8838: Because `Target.get_facts` raised for a level the machine could not answer about, and there was no non-throwing query. That refusal is now gone: a target that never said how many cards is one card, which is a true answer, so `parallel_capacity` answers for every level and the probe had nothing left to catch. `_coarsest_measured_level` is deleted and `level` falls back to `topologies[0].name`.

src/tilefoundry/runtime/decorator.py
zhen8838: It should not have been. It asked torch for the process current device while the resource read on whatever `device=` it was constructed with, so `"cpu"` or `"cuda:3"` made the cache key describe something other than where the data was.

The root cause was one level up: `Placement` was a callable and every caller read the current device from it, so a loaded module answered differently on each call and the device had to be in the key to tell those answers apart. `Placement` is now the mapping `{level: id}`, fixed at load; the key is the weight name, and `_current_device` is deleted.

src/tilefoundry/runtime/decorator.py
zhen8838: It is one hook, in and out, and nothing said so. `_call` in `_make_kernel_caller` is the hook; its docstring now names the way in as the `forward_pre_hook` side and the way out as the `forward_hook` side, after `distribute_module(input_fn=, output_fn=)`.

`_unwrap` and `_wrap` are `_to_local` and `_from_local`, after `DTensor.to_local` / `from_local` -- the same question, and PyTorch own words for it. `_program` is gone with the cache key.

src/tilefoundry/runtime/tensor.py
zhen8838: `_one_of` narrowed one tensor axis to one part of it and `_check` held the result against `local_type_of`. Both are gone: the loop they sat in was a second implementation of the device side `local_layout_and_offset`, and `_check` existed only because the two could disagree.

There is one implementation now, `local_window(shard, tensor_shape, ids)` in `ir/types/shard/local.py`, answering per tensor axis the `(start, extent)` this instance holds -- the same rule the C++ side runs (connected division, outermost first). `to_local` takes one basic slice from it. The agreement `_check` guarded is a property of two implementations, so it is `tests/ir/types/test_local_window.py` now, which also holds that every instance windows tile the tensor exactly once.

src/tilefoundry/target/cuda/facts.py
zhen8838: Because the peak being divided was one card, so a gpu unit -- one card -- divided it by one. The name said "per device" while the numerator was also per device, which is what made 1 look like a missing `device_count`.

You are right that the total should be the deployment. The peaks now start there (`card_peak * device_count`), so one function `parallel_units` answers how many of a level run at once: `gpu -> device_count`, `cta -> cards * sm_count`, `thread -> cards * threads_per_card`. It is both the divisor for a per-unit rate and the parallel capacity, so `parallel_capacity` three-branch `if` is one line. Per-unit answers are unchanged; `ThroughputFacts` is scaled to match so both halves of a roofline start from the same total.

tests/fixtures/placed/tp_all_to_all.py
zhen8838: It is a fixture, not a test: `tests/analysis/test_analysis_families.py` uses it to hold that a `Reshard` across a gpu mesh axis reports `SENT_BYTES` crossing per card, and that a unit inside that boundary states no share of it. The module docstring now says so.

tests/fixtures/placed/tiny_tp_decoder.py
zhen8838: It does check them, and per device: `PROJECT_SHARDS = {gpu: PROJECT_FULL[:, gpu*COLUMNS:(gpu+1)*COLUMNS]}` gives each card a different column block, and the twin body asserts `torch.equal(project_weight.cpu(), PROJECT_SHARDS[...])`. Two cards therefore assert two different tensors, not the same one twice.

What did change is where the card comes from. It read `torch.cuda.current_device()`, which only worked because one module was being moved between cards; the test now builds one module per card in its own thread, so the twin is handed the card its `Placement` names.

docs/spec/runtime.md
```diff
@@ -12,14 +12,14 @@ snapshot.
 
 ## 1. Python Runtime Surface
 
-### 1.1 `RuntimeModule`
+### 1.1 `runtime/module.py`
```
zhen8838, line 15: Organizing chapters according to files does not mean replacing chapter names with file names. Each file should have its own independent classes and functions. Chapters are still named by classes and functions, and the code organization should be reversely optimized in this way.

docs/spec/target.md
```diff
@@ -244,8 +244,15 @@ hierarchy stops at those levels; warp, lane, and warpgroup structure belongs in
 thread mesh layouts.
 
 - constraints:
-  - `CudaTarget.topology_levels` MUST be `("cta", "thread")` for this
-    single-device target.
+  - `CudaTarget.topology_levels` MUST be `("gpu", "cta", "thread")`, outermost
```
zhen8838, line 247: The details of this tritival do not need to be explained. A literal [str] can be made in the code, so that the restriction can be clear at a glance, and there is no need to repeat the spec explanation.

include/tilefoundry/runtime/cuda/ops/copy/copy_impl.h
```diff
@@ -8,8 +8,8 @@ namespace copy_impl {
 struct Copy {
     template <class TSrc, class TDst>
     __device__ void operator()(TSrc const &src, TDst &dst) const {
-        auto &&s = detail::to_local(src);
-        auto &&d = detail::to_local(dst);
+        auto &&s = detail::local_tensor(src);
```
zhen8838, line 11: Local_tensor cannot be put into detial.

include/tilefoundry/runtime/cuda/runtime.cuh
```diff
@@ -56,18 +57,96 @@ CUTE_HOST_DEVICE constexpr auto dims_of() noexcept {
 /// extents its own multiply up to, and a stride counts the levels below it.
 template <TopologyScope T>
 CUTE_HOST_DEVICE constexpr auto program_shape() noexcept {
```
zhen8838, line 59: This is not written correctly. It should be expanded directly from T to scope count.

include/tilefoundry/runtime/cuda/runtime.cuh
```diff
@@ -56,18 +57,96 @@ CUTE_HOST_DEVICE constexpr auto dims_of() noexcept {
...
+///
+/// A translation unit states it by defining TILEFOUNDRY_PROGRAM_LEVEL before
+/// this header, the way the target macro selects a runtime. A macro rather
+/// than a function the .cu defines, because the levels decide how long a
+/// coord is and a length is read while this header is compiled.
+#if !defined(TILEFOUNDRY_PROGRAM_LEVEL)
+#define TILEFOUNDRY_PROGRAM_LEVEL tilefoundry::TopologyScope::cta
+#endif
+
+CUTE_HOST_DEVICE constexpr TopologyScope program_level() noexcept {
```
zhen8838, line 80: It is more reasonable to change program_level to program_topoogy.  Other macros are also renamed accordingly.

src/tilefoundry/analysis/check.py
```diff
@@ -122,8 +122,8 @@ class PerformanceChecker:
     def check_target_facts(self, ctx: AnalysisCheckContext) -> None:
         """Require a machine whose stated capacity and rates fit the question."""
         try:
-            capacity = ctx.target.get_facts(ParallelCapacityFacts)
-            services = ctx.target.get_facts(PerformanceServiceFacts)
+            capacity = ctx.target.get_facts(ParallelCapacityFacts, ctx.level)
```
zhen8838, line 125: It should be ctx.current_topology

src/tilefoundry/cli/analyze.py
```diff
@@ -83,10 +82,10 @@ def guidance() -> str:
 
         family         what --topology changes                 pass it when
         ------------   --------------------------------------  ---------------------
-        compute-cost   flops_per_unit and service_per_unit.    the program shards
-                       flops and service stay global
-        memory         per-unit traffic. Footprint follows     the program shards
-                       its owner for each storage level
+        compute-cost   nothing. Every kind states its total     never
```
zhen8838, line 85: I hope that the naming of the current topology of unified runtime and Python should not be called level.

src/tilefoundry/codegen/cpu/module.py
```diff
@@ -161,11 +161,27 @@ def _placement_line(name: str, storage) -> str:
     )
 
 
-def _shim_decl(fn: PrimFunction) -> str:
+_GPU_ID_PARAM = "tf_gpu_program_id"
```
zhen8838, line 164: This tag name is unreasonable. Please refer to how other repos build exclusive tags.

src/tilefoundry/codegen/cpu/module.py
```diff
@@ -161,11 +161,27 @@ def _placement_line(name: str, storage) -> str:
...
+
+
+def _places_gpu(module: Module) -> bool:
```
zhen8838, line 167: Why does such a reverse dependence problem occur?

src/tilefoundry/ir/types/shard/local.py
```diff
@@ -0,0 +1,121 @@
...
+                )
+            held //= count
+            start += positions[mesh_axis] * held * _inner(
+                shard, tensor_shape, mesh_axis, axis
+            )
+        window.append((start, held))
+    return tuple(window)
+
+
+__all__ = ["divisor", "local_window"]
```
zhen8838, line 121: Why can't this function be consistent with the runtime implementation, naming and logic?

include/tilefoundry/runtime/cuda/ops/copy/copy_impl.h
zhen8838: Done. `local_tensor`, `local_view_t` and `ShardTensorLike` are now `tilefoundry::` names; the ops call them unqualified. Only `is_shard_tensor` — how `ShardTensorLike` is decided, which nothing outside needs — stays in `detail`. A step every op must take is not an implementation detail, so `ops::detail` now holds only implementation names an op *may* want and none must have.

include/tilefoundry/runtime/cuda/runtime.cuh
zhen8838: Done. `program_shape<T>()` now expands `topologies_from<T>()` — the run from `T` to `scope_count` — and maps `program_dim` over it with `cute::transform`, instead of the hand-written recursion.

include/tilefoundry/runtime/cuda/runtime.cuh
zhen8838: Done. `program_level` → `program_topology`, the macro `TILEFOUNDRY_PROGRAM_LEVEL` → `TILEFOUNDRY_PROGRAM_TOPOLOGY`, and the Python side follows (`analysis` / `codegen` / `ir` all say topology now). Two hierarchies shared the word "level" — execution and memory — and each now says which it is.

src/tilefoundry/codegen/cpu/module.py
zhen8838: Done. Every identifier codegen invents now lives in `codegen/names.py` under one `tilefoundry_` prefix — which clears both the names a user writes in a kernel and the ones C++ reserves (leading double underscore):

```
tilefoundry_<name>_host      was __tilefoundry_<name>_host
tilefoundry_<name>_launch
tilefoundry_<name>_kernel    was <name>_kernel, which carried a raw `$` for a mangled variant
tilefoundry_gpu_program_id   was tf_gpu_program_id / gpu_program_id, inconsistent between shim and host
tilefoundry_meta             was meta
```

A user's own parameter names are never prefixed.

src/tilefoundry/codegen/cpu/module.py
zhen8838: It came from the ABI types living in `runtime/function.py` while their producer is codegen — `codegen/linker.py` was importing `EntryABI` from the runtime.

Fixed by moving them where they are made: the whole family is now `codegen/signature.py`, and the runtime imports it. Two more reverse edges went with it — `codegen/topology.py` no longer imports `runtime.function` (`places_of` moved next to `places_any`), and `codegen/cpu/module.py` no longer imports `codegen/cuda/module.py` (that import existed only for `shim_symbol`, now `names.launch_shim`).

src/tilefoundry/ir/types/shard/local.py
zhen8838: Done — it is now the same algebra under the same name on both sides.

`local_window` is gone. Host and device both call `local_layout_and_offset(shard, shape, ids) -> (Layout, offset)`: the host slices with `as_strided`, the device with `make_tensor(engine.data() + off, layout)`. One layout and one offset, derived once.

It was also wrong. The old `local_window` multiplied each start by the *running* `held` rather than the final local extent, so for a `(32,)` tensor over a 4x8 mesh it produced starts `[0..7, 64..71, 128..135, 192..199]` — three of the four past the end of the tensor. My own tiling test could not catch it: it counted distinct points, not in-bounds ones.

docs/spec/target.md
zhen8838: Done. `topology_levels` is gone as a ClassVar; the tuple is now stated once as a literal in the code that owns it:

```python
# src/tilefoundry/target/cuda/target.py
return TopologyFacts((
    TopologyLimitFacts("gpu", self.device_count, from_target=True),
    ...
))
```

and the spec paragraph shrank from three and a half lines restating the code to one line stating the rule the code does not say by itself:

> A topology level whose `from_target` is set takes its extent and its program ids from the target instance rather than from the hardware document or a register.

docs/spec/runtime.md
zhen8838: Landed as a compromise between this and what you asked for later ("像 C++ 那边一样按文件夹和文件名来组织"): §1 is one **section** per file, and one **heading** per class under it.

```
### 1.1 `runtime/module.py`
#### `RuntimeModule`
#### `CompiledModule`      <- was §1.5 "Calling convention"
### 1.2 `runtime/function.py`
### 1.3 `runtime/decorator.py`
#### `@runtime_module` / `@runtime_func`
### 1.4 `runtime/loader.py`
#### Launcher ABI          <- was §1.7
```

So a class still has its own named heading, and the section says which file owns it — the same shape §2 already had for the headers.

The code was reverse-optimized the way you asked: §1.5 and §1.7 were sections that named no file, and they are now headings under the file that owns them. Their numbers are left as gaps rather than renumbering the rest, per `docs/SPEC-RULES.md`: "Append a section; do not renumber one."

Every block is now read out of the file by `scripts/runtime_spec_surface.py` (Python via `ast`, C++ via libclang) instead of hand-copied, and a new pre-commit hook checks it.

src/tilefoundry/analysis/check.py
zhen8838: Done, as `ctx.topology_level` rather than `ctx.current_topology`.

The ambiguity was not that "level" is a bad word — it is that two hierarchies shared it: execution (`gpu`/`cta`/`thread`) and memory (`gmem`/`smem`/`rmem`). So each carries its own prefix instead of one of them getting a new word:

```
ctx.level / analyze(level=)      ->  ctx.topology_level / analyze(topology_level=)
local_type_of(level=)            ->  local_type_of(topology_level=)
LevelFootprint.level, the        ->  memory_level
  TrafficMetadata.storage key,
  MemoryHierarchyFacts
TILEFOUNDRY_PROGRAM_LEVEL        ->  TILEFOUNDRY_PROGRAM_TOPOLOGY
program_level()                  ->  program_topology()
```

Runtime and Python now say the same word for the same hierarchy. Where `topology` is already the subject the `level` is dropped, since it would say the same thing twice: `level_axes` -> `topology_axes`, `level_projection` -> `topology_projection`. It stays in a parameter name, where it does mean *which* one.

src/tilefoundry/cli/analyze.py
zhen8838: Done, as `ctx.topology_level` rather than `ctx.current_topology`.

The ambiguity was not that "level" is a bad word — it is that two hierarchies shared it: execution (`gpu`/`cta`/`thread`) and memory (`gmem`/`smem`/`rmem`). So each carries its own prefix instead of one of them getting a new word:

```
ctx.level / analyze(level=)      ->  ctx.topology_level / analyze(topology_level=)
local_type_of(level=)            ->  local_type_of(topology_level=)
LevelFootprint.level, the        ->  memory_level
  TrafficMetadata.storage key,
  MemoryHierarchyFacts
TILEFOUNDRY_PROGRAM_LEVEL        ->  TILEFOUNDRY_PROGRAM_TOPOLOGY
program_level()                  ->  program_topology()
```

Runtime and Python now say the same word for the same hierarchy. Where `topology` is already the subject the `level` is dropped, since it would say the same thing twice: `level_axes` -> `topology_axes`, `level_projection` -> `topology_projection`. It stays in a parameter name, where it does mean *which* one.

src/tilefoundry/codegen/cuda/module.py
```diff
@@ -8,58 +8,99 @@
...
+from tilefoundry.codegen import names
 from tilefoundry.codegen.cuda.context import CodegenContext
 from tilefoundry.codegen.cuda.templates import render
 from tilefoundry.codegen.cuda.tir.prim_function import (
     _compute_kernel_fields,
 )
 from tilefoundry.codegen.linkable import LinkableFunction, LinkableModule
 from tilefoundry.codegen.registry import CodeGenerator
+from tilefoundry.codegen.signature import (
+    GPU_ID,
```
zhen8838: Why does code gen specifically put out the GPU ID here?

src/tilefoundry/codegen/cuda/module.py
```diff
@@ -8,58 +8,99 @@
...
+        if fields.param_kinds[param.name] != "tensor":
+            return "int"
+        return f"{fields.param_cpp_types[param.name]}*"
+
+    return ctype
 
 
-def _emit_kernel_and_shim(fields) -> str:
-    shim_params = []
+def _emit_kernel_and_shim(fields, places_gpu: bool = False) -> str:
```
zhen8838: I feel that the logic of code generation here still needs to be reorganized.

src/tilefoundry/codegen/cuda/tir/prim_function.py
```diff
@@ -10,6 +10,7 @@
 
 from dataclasses import dataclass
 
+from tilefoundry.codegen import names
```
zhen8838: Names is also very strange.

src/tilefoundry/compile.py
zhen8838, line 112: Why do you need to import so much content in one function?

src/tilefoundry/target/cuda/target.py
```diff
@@ -100,6 +117,7 @@ def __init__(
             )
         if architecture_id is not None and device_id is not None:
             check_compatible(architecture, device)
+        object.__setattr__(self, "device_count", device_count)
```
zhen8838, line 120: Why object setattr? Why is this class frozen?

tests/analysis/test_analysis_families.py
```diff
@@ -442,5 +447,83 @@ def test_a_price_is_refused_where_the_machine_states_no_rate_to_pay_it_at() -> N
             throughput,
             replace(services, unit_bandwidth=()),
             moved=crossed,
-            level="cta",
+            topology_level="cta",
         )
+
+
+def test_a_level_the_machine_can_answer_about_is_one_it_measures() -> None:
```
zhen8838: There is no need to add so many conservative tests. You only need to add a test to monitor whether the analysis results of the current program are desired. The value of expected can be written directly.

src/tilefoundry/codegen/cuda/module.py
zhen8838: It does not any more. The emitter never names `gpu`.

Which levels a host must supply ids for is `TopologyFacts.from_target` — a level no device register answers for. That question had one honest answer and four places asking it, and the answer was being collapsed to a bool (`places_any`), which left the emitter no option but to write the level name in by hand: `GPU_ID`, `places_gpu`, `TopologyScope::gpu`, `places_any`.

Now `program_id_params(module, target)` asks once per compile and hands down one `ProgramIdSignature` per level. The emitter writes one line per signature, naming none of them:

```python
f"{META.name}.program_id[int(...TopologyScope::{p.topology_level})] = static_cast<int>({p.name});"
```

`grep -rn gpu src/tilefoundry/codegen/` finds nothing.

It also closed a real gap: the C++ side declared exactly one such parameter while the loader passed one per level. Only `gpu` sets `from_target` today, so they agreed — a second one would not have been caught at compile time.

src/tilefoundry/codegen/cuda/module.py
zhen8838: Reorganized, and the thing that was in the way is gone.

The emitter is the visitor that walks a body. In front of it sat `_KernelFields`: fourteen precomputed fields, six with no reader left, described in its own docstring as shared with a single-source emitter that no longer exists. Then a template rendered from ten keyword arguments, with the `ProgramMetaData` statement built as a Python string carrying an embedded newline and spliced into a list of casts — a second rendering point outside the template.

Now a compile settles every signature first, and two passes over each function write its declaration and its definition, the second by walking the body. A translation unit is assembled from its functions rather than rendered a second time beside them: `LinkableModule.source` is the one assembly point, so a forward declaration and its definition come from one `CallableSignature` and cannot drift.

Two defects surfaced on the way. `program_dim<>` is specialized once per translation unit, and the old body-walk multiplied the `gpu` level into the thread count — a placed kernel believed it had 32 threads while the launch gave it 16. And two host entries both named `main` shared a build directory, so the second compile dlopened the first one's library.

src/tilefoundry/codegen/cuda/tir/prim_function.py
zhen8838: Deleted. `codegen/names.py` is gone.

The name was strange because the module was: it was nobody's names. How a symbol is spelled is the target language's business — `tilefoundry_x_launch` is a CUDA shim convention, `tilefoundry_x_host` a host-ABI one — and a neutral module spelling them for every target is one place deciding something it does not own. Renaming it would have kept that.

They now live beside the handlers of the target that spells them: `codegen/cuda/abi.py` and `codegen/cpu/abi.py`, registered against `(target, role, type)` like everything else codegen dispatches on.

The same sweep took the last of the old prefix: `tf_grid_bar_state`, `tf_ew_*`, `tf_copy_*`, `tf_mbar_ready`, `__tf_dispatch_subject`, and `tilefoundry::tf_grid_bar_state` which carried a namespace and a prefix at once. `grep -rn "tf_" src/tilefoundry/codegen/` is empty.

src/tilefoundry/compile.py
zhen8838: The block is gone — all of it, and there is no comment in its place.

The claim it rested on ("keep these heavy codegen/runtime imports off the module load") turned out to be false, so rather than argue about it I measured:

| | before | after |
|---|---|---|
| `import tilefoundry` (3 runs) | 1.359 / 1.365 / 1.377 s | 1.366 / 1.372 / 1.388 s |
| modules in `sys.modules` | 1463 | 1515 |
| **third-party / stdlib modules added** | — | **0** |

Hoisting everything adds 52 pure-`tilefoundry` modules and not one new third-party or stdlib import — about 10 ms on a 1.37 s import. The genuinely heavy things are either already loaded before `compile.py` is reached (`torch`, `numpy`, `isl`) or already lazy one level down where they belong: `runtime/loader.py` imports `tvm_ffi` inside `load_linked_module`, and `codegen/linker.py` shells out to CMake instead of importing a toolchain. `tilefoundry.target` was a module-level import in the same file, so `CpuTarget` / `CudaTarget` were lazy for no reason at all.

`compile.py` now has zero function-level imports and zero `noqa: PLC0415`.

src/tilefoundry/target/cuda/target.py
zhen8838: Not changed this round, deliberately — I want to say why rather than leave it looking overlooked.

`frozen=True` has a stated reason, in the base class: "A target is a value: what it knows is its hardware ... two equal targets are interchangeable everywhere." `init=False` has one too: construction parses and cross-checks. `device` may be a `Device`, a string or a `Path`; `architecture` may be omitted and derived from the device; an explicit `arch` that disagrees with `architecture.name` is refused. A generated `__init__` can do none of that, so the nine `object.__setattr__` calls are what a hand-written `__init__` costs on a frozen dataclass. `AmxTarget` is the same shape with eight.

So the question worth asking is not "why setattr" but "should parsing and validation happen in the constructor or in a classmethod factory that builds the validated values and hands them to a generated `__init__`" — which would leave zero `object.__setattr__` and keep the value semantics. That is a change to how every Target is constructed, and it has nothing to do with codegen, which is what this round was about. Raising it as its own issue rather than folding it into a PR you already asked to narrow.

This round touched `CudaTarget` only to add `device_count` and the `TopologyFacts` projection that reads it.

tests/analysis/test_analysis_families.py
zhen8838: Collapsed to one, with the values written in.

You were right about what was wrong with them, and it is worse than "too many": two of the three asserted *relations*, not numbers.

```python
assert work["cta"] == work["thread"]
assert fine <= coarse and coarse % max(fine, 1) == 0
```

Both hold for values that are wrong. Break the analysis and they stay green.

There is now one test that runs the analysis at `cta` and at `thread` and compares each against a literal dict — `whole_program`, `per_cta`, `per_thread`, `end_ns`, the levels it reports, the level it answered at. The numbers came from running it and reading what it reported. `test_a_reshard_that_changes_shards_says_what_left_the_card` already asserted values computed from named constants, so it stayed.

One thing that pinning the numbers exposed, which I have not fixed: `whole_program` for f32 is 128 when asked at `cta` and 2112 when asked at `thread`, while `per_cta` and `per_thread` do not move. `Spread`'s own docstring says "One quantity, whole and as one unit of each topology level holds it" — the whole of a quantity should not depend on which level you asked about. It is not new (`test_scope_positions_turn_per_unit_cost_into_total_cost` already pins `(512, 128)`, so unrolling replicated work into the total at a finer level is deliberate and tested), and changing it is an analysis decision, not a codegen one. The number is pinned and the docstring says so, so a future fix shows up as a test someone chose to change rather than as silent drift.

PR #166, opened 2026-09-13
feat(inspection,parser): print placed types as parser sugar

zhen8838:
## Why
- PR #166's review: the printer dispatched types through `render_value()` instead of visiting each expression's type; `render_*` methods shadowed `visit_*`; the goldens showed canonical `ShardLayout(...)` dataclasses instead of parser sugar; the fixtures were too simple to show a composed mesh, a broadcast and a partial together; and the work had grown four tests where two were asked for.
- The sugar the printer already emitted was never readable back: `((dims), (strides))` and `{axis @ P("sum")}` had no parser production at all, and a bare `(8,)` reparsed as a plain `Layout` that dropped the mesh.

## What
- `PythonPrinter` inherits `PythonTypePrinter`. Every type has exactly one `visit_<Type>`, expression printers emit `expr.type` through `self.visit`, and the `render_*` parallel entries and `owner` back-delegation are gone.
- One parser production covers the whole layout sugar — dims with each split written on the dimension it divides, an optional stride tuple, and an optional `{axis @ B(), axis @ P("sum")}` set. `ExplicitLayoutPattern` folds into it.
- Every shard layout over a named mesh prints as that sugar, in type slots and in op-attribute slots. The TIR fixtures are authored in it and print back to their own source.
- Structurally identical meshes share one prelude name, and a printed type names only a mesh the prelude binds.
- Withdrew PR #166's footprint-by-level projection, its `details` opt-in, and the spec text describing them, along with the fixtures and tests that existed only to pin that projection.

## Contract
- `docs/spec/inspection.md` §2.3 states when the printer MUST use placement sugar and how it writes splits, strides and value states; §2.5 states one name per mesh descriptor and no named mesh the prelude does not bind.
- `docs/spec/parser.md`'s generated grammar: `placed-layout` now carries the stride tuple and the value-state set; `explicit-layout` is gone.
- An unstated stride tuple stays the sugar default a `Reshard` materializes from the storage tier it moves between; the sugar states strides whenever the layout has them.
- `TensorView.layout` is annotated `LayoutBase`, so its slot parses as a layout.
- `tests/inspection/test_tir_roundtrip.py` asserts fixture source equality again, as inspection §2.7 requires.

## Risk
- `tests/fixtures/tir/mma.py` uses placement sugar with named `warp/lane` axes; the CUDA MMA tile scope uses the same mesh names so atom fragment layouts remain structurally equal.
- Annotations are longer now that they state strides. That is what makes an annotation name the same layout the emitted code does.


docs/spec/analysis.md
```diff
@@ -304,7 +279,10 @@ class LoopFootprintMetadata(IRMetadata):
     """Known buffer accesses or a lower bound within one authored LoopRegion.
 
     Attributes:
-        footprints: attribute; One row per source buffer and storage level.
+        footprints: attribute; One row per source buffer and storage level. The
```
zhen8838: Why do you need to modify the description of this?

docs/spec/inspection.md
```diff
@@ -205,6 +205,11 @@ still renders verbose, so no annotation loses information. The annotation is
 **display-only** ([§2.7](#27-round-trip-contract)); what round-trips is the
 emitted code, not its comments.
 
+All canonical type values are dispatched through the shared `TypeFunctor` /
+`PythonTypePrinter` implementation. HIR and TIR retain their own function and
+statement printers, but `render_mode()` MUST NOT change the syntax of a
```
zhen8838: Why do you have to write render mode here? Shouldn't there be this thing?

src/tilefoundry/inspection/printer_base.py
```diff
@@ -12,51 +12,43 @@
...
+        mesh_name = ctx.mesh_alias(value.mesh) if ctx is not None else None
+        if mesh_name is None or not value.mesh.names:
+            return None
+        count = ctx.mesh_count() if ctx is not None and hasattr(ctx, "mesh_count") else 1
+        return _shard_layout_surface_str(value, mesh_name=mesh_name, mesh_unique=count == 1)
 
     def atom_reference(self, value, ctx=None) -> str:
         return f"T.cuda.mma.atom(op=T.cuda.mma.{value.op.name})"
 
     def render_value(self, value, ctx=None, indent: str = "") -> str:
```
zhen8838, line 347: Why is the call process render value? It should be to visit each expression, and then visit the type of this expression.

src/tilefoundry/inspection/printer_base.py
```diff
@@ -85,133 +77,28 @@ def shape_tuple(self, shape: tuple, ctx=None) -> str:
...
-            if surface is not None:
-                result += f", {surface}"
-            else:
-                result += f",\n{indent}    {self.render_shard_layout(ty.layout, ctx, indent + '    ')}"
-        if ty.storage is not StorageKind.GMEM:
-            result += f', "{ty.storage.name.lower()}"'
-        return result + "]"
+        return self.type_printer.render_tensor_type(ty, ctx, indent, is_const)
 
     def _render_layout_positional(self, layout, ctx=None):
```
zhen8838: All the _render_ here are not correct and need to be corrected.

src/tilefoundry/inspection/python_type_printer.py
```diff
@@ -0,0 +1,202 @@
...
+            f'Topology("{topology.name}", {self.owner.dim_entry(topology.size, ctx)})'
+            for topology in mesh.topologies
+        )
+        topologies = f"({values}{',' if len(mesh.topologies) == 1 else ''})"
+        result = f"Mesh({topologies}, {self.render_layout(mesh.layout, ctx, indent)}"
+        if mesh.names:
+            result += f", names={tuple(mesh.names)!r}"
+        return result + ")"
+
+    def render_shard_layout(
```
zhen8838: Why shouldn't render_xxx use Visit_xxx directly?

tests/fixtures/hir/tensor_types.py
```diff
@@ -0,0 +1,58 @@
...
+@module(
+    entry="loop_nest",
+    target=CudaTarget("nvidia.h200_sxm"),
+    topologies=(Topology("cta", 1), Topology("thread", 4)),
+)
+class TensorTypes:
+    @func
+    def broadcast(x: Tensor[(8,), "f32"]):
+        with Mesh(("thread",), (4,), ("lane",)) as _m:
+            return tf.reshard(x, (8,), "gmem")
```
zhen8838: You didn't test what I wanted. I want to test something similar. Tensor[((? @ m.x, ? @ m.y), { ? @ B, ? @P})]

tests/fixtures/hir/tensor_types.py
```diff
@@ -0,0 +1,58 @@
...
+
+    @func
+    def split_1d(x: Tensor[(16,), "f32"]):
+        with Mesh(("thread",), (4,), ("lane",)) as m:
+            local = tf.reshard(x, (16 @ m.lane,), "rmem")
+            return tf.reshard(local, (16,), "gmem")
+
+    @func
+    def split_2d(x: Tensor[(8, 16), "f32"]):
```
zhen8838: All the examples here must be complicated enough. You are too simple.

tests/inspection/golden/tensor_types.loop_nest.analyze.golden
```diff
@@ -0,0 +1,29 @@
...
+from tilefoundry.ir.types.shard import B, Layout, Mesh, S, ShardLayout, Topology
+
+thread = Mesh((Topology("thread", 4),), Layout((4,), (1,)), names=('lane',))
+
+@func
+def loop_nest(
+    x: Tensor[(8,), "f32"]
+) -> Tensor[(8,), "f32", (8,)]:
+    with thread as _thread:  # Tensor[(8,), "f32", (8,)]
+        carried = reshard(x, layout=ShardLayout(
```
zhen8838: Didn't I say that the printout must be grammar candy? Why are you still in the form of canonical layout?

tests/inspection/golden/tensor_types.loop_nest.golden
```diff
@@ -0,0 +1,29 @@
+from __future__ import annotations
+
+from tilefoundry import func
+from tilefoundry.dsl.tf import *  # noqa: F401, F403
+from tilefoundry.dsl import Tensor
+from tilefoundry.dsl.storage import gmem, host, rmem, smem, tmem  # noqa: F401
+from tilefoundry.ir.types.shard import B, Layout, Mesh, S, ShardLayout, Topology
+
+thread = Mesh((Topology("thread", 4),), Layout((4,), (1,)), names=('lane',))
```
zhen8838: Why is your golden test input not the complicated example above? Instead, you used another very simple example.

tests/inspection/golden/tensor_types.loop_nest.analyze.golden
```diff
@@ -0,0 +1,29 @@
...
+                layout=Layout((4, 2), None),
+                attrs=(S(0),),
+                mesh=thread,
+            ), storage=rmem)  # Tensor[(8,), "f32", ((4 @ thread.lane, 2), (0, 1)), "rmem"]; compute-cost; traffic traffic=gmem:r32/w0@r32/w0,rmem:r0/w32@r0/w32
+        for _ in range(2):  # Tensor[(8,), "f32", ((4 @ thread.lane, 2), (0, 1)), "rmem"]; loop-footprint footprints=rmem:64 status=complete
+            v1 = unary(carried, kind="square")  # Tensor[(8,), "f32", ((4 @ thread.lane, 2), (0, 1)), "rmem"]; compute-cost flops=f32:8@8; traffic traffic=rmem:r32/w32@r32/w32
```
zhen8838: Then there is no broadcast and partial in all the tensor types of your analyze here, but that's the situation I want to test.

tests/inspection/test_analyze_render_golden.py
```diff
@@ -0,0 +1,38 @@
...
+
+from tests.fixtures.hir.tensor_types import TensorTypes
+from tilefoundry.analysis import analyze
+from tilefoundry.analysis.metadata import BufferFootprint, LoopFootprintMetadata
+from tilefoundry.inspection import as_script
+from tilefoundry.inspection.analysis_report import render_analysis
+from tilefoundry.inspection.values import render_comment
+
+
+def test_loop_footprint_comment_aggregates_by_level() -> None:
```
zhen8838: I said that only two tests will be added to the whole work. One is the golden of round trip. One is the golden of analyze.

docs/spec/analysis.md
zhen8838: Reverted. The by-level aggregation and the `details` opt-in that description was written for are gone, so `LoopFootprintMetadata.footprints` is back to what it said before.

docs/spec/inspection.md
zhen8838: Removed. `render_mode()` itself is gone from `PrintContext` too: now that HIR and TIR share one type visitor, nothing reads it, so there is no mode that could change a type's syntax.

src/tilefoundry/inspection/printer_base.py
zhen8838: Changed. `PythonPrinter` now inherits `PythonTypePrinter`, so a printer holding an `expr.type` writes `self.visit(expr.type, ctx)` — the same `visit` that emits the type's children. `render_value()` is left only for op-attribute literals that are not part of the type language (`MmaAtom`, enums, `Target`, plain constants).

src/tilefoundry/inspection/printer_base.py
zhen8838: All of them are gone, and the `owner` back-delegation with them. `printer_base.py` now holds only `render_value`, `render_pattern` and `atom_reference`.

src/tilefoundry/inspection/python_type_printer.py
zhen8838: Folded. `render_tensor_type` / `render_layout` / `render_mesh` / `render_shard_layout` / `_shard_attr_str` are now the bodies of `visit_TensorType` / `visit_Layout` / `visit_Mesh` / `visit_ShardLayout` / `visit_Broadcast|Split|Partial`, and they recurse into each other through `self.visit`.

tests/fixtures/hir/tensor_types.py
zhen8838: That form now parses and prints. The parser had no production for the value-state set at all, and none for a stride tuple beside a placement — one production now covers dims + optional strides + `{axis @ B(), axis @ P("sum")}`. See `tests/fixtures/inspection/type_printer_sugar.py`: `acc` is split + partial, `weight` is broadcast + partial, `mixed` is split + broadcast + partial across a composed mesh.

tests/fixtures/hir/tensor_types.py
zhen8838: Replaced. `tests/fixtures/hir/tensor_types.py` is deleted. `tests/fixtures/inspection/type_printer_sugar.py` has two functions that carry a 1-D outer mesh and a 2-D inner mesh composed on one value, splits on one and on three tensor axes, explicit strides, rank expansion, two dtypes, gmem/rmem/smem, and a loop whose two carried fields are placed differently.

tests/inspection/golden/tensor_types.loop_nest.analyze.golden
zhen8838: Done. Every shard layout over a mesh with named axes prints as sugar now, in type slots and in op-attribute slots (`reshard(..., layout=(4 @ cta.tile, 2, ...))`). The round-trip test asserts `ShardLayout(` does not appear in the golden at all.

tests/inspection/golden/tensor_types.loop_nest.golden
zhen8838: Both goldens read the same fixture now: `tests/fixtures/inspection/type_printer_sugar.printed.txt` and `.analyzed.txt`. The analyze test also asserts that stripping the metadata comments gives back the round-trip text, so the two cannot drift.

tests/inspection/golden/tensor_types.loop_nest.analyze.golden
zhen8838: Both are there: `{thread.lane @ P("sum")}` and `{cta.lane @ P("sum")}` on the parameters, `{thread.warp @ B(), thread.lane @ B()}` on the results, and the loop carries one split field beside one all-broadcast field.

tests/inspection/test_analyze_render_golden.py
zhen8838: Two, and only two: `test_placed_types_print_and_reparse_as_layout_sugar` (round-trip golden) and `test_analysis_annotates_the_printed_program_without_changing_it` (analyze golden). `test_type_printer_golden.py` and `test_analyze_render_golden.py` are deleted; `git diff main -- tests/ | grep '^+def test_'` prints those two lines.

docs/spec/parser.md
```diff
@@ -94,9 +94,23 @@ dim-expr              ::= integer-literal
                           | dim-expr ('+' | '-' | '*' | '//' | '%') dim-expr
                           | (identifier | primary '.' identifier) '(' (dim-expr (',' dim-expr)*)?
                             ')'
-placed-shape          ::= '(' ((expression '@' ('(' mesh-axis (',' mesh-axis)* ')' | mesh-axis) |
-                          dim-expr) (',' (expression '@' ('(' mesh-axis (',' mesh-axis)* ')' |
-                          mesh-axis) | dim-expr))*)? ')'
+placed-shape          ::= '(' '(' ((expression '@' ('(' mesh-axis (',' mesh-axis)* ')' | mesh-axis)
```
zhen8838: Why has this pattern become so complicated? And I see that every line seems to be the same before.

docs/spec/parser.md
```diff
@@ -133,15 +147,27 @@ expression            ::= literal
                           | subscript
 call                  ::= expression '(' ((expression | keyword-name '=' expression) (','
                           (expression | keyword-name '=' expression))*)? ')'
-explicit-layout       ::= '(' (tensor-shape-layout | shape) ',' shape ')'
-placed-layout         ::= '(' ((expression '@' ('(' mesh-axis (',' mesh-axis)* ')' | mesh-axis) |
-                          dim-expr) (',' (expression '@' ('(' mesh-axis (',' mesh-axis)* ')' |
-                          mesh-axis) | dim-expr))*)? ')'
+placed-layout         ::= '(' '(' ((expression '@' ('(' mesh-axis (',' mesh-axis)* ')' | mesh-axis)
```
zhen8838: Why does this and place shape coincide highly?

src/tilefoundry/inspection/python_printer.py
```diff
@@ -274,13 +265,13 @@ def _ceildiv_args(entry: Call) -> tuple[object, object] | None:
 
 def _classify_shard_attrs(
```
zhen8838: Why is this function still placed here in Python print instead of python type printer?

src/tilefoundry/inspection/python_printer.py
```diff
@@ -298,39 +290,36 @@ def _classify_shard_attrs(
...
+    return splits, partials, broadcasts
 
 
-def _shard_layout_surface_str(
-    sl: ShardLayout,
-    mesh_name: str = "gpu",
-    *,
-    mesh_unique: bool = False,
-) -> str | None:
+def _shard_layout_surface_str(sl: ShardLayout, mesh_name: str = "gpu", ctx=None) -> str | None:
```
zhen8838: Why is this string not visited, but by such a function?

src/tilefoundry/inspection/python_printer.py
```diff
@@ -1292,23 +1284,27 @@ def _shape_tuple(shape: tuple) -> str:
     return _HIR_RENDERER.shape_tuple(shape)
 
 
+def _type_str(value, ctx=None, indent: str = "", *, is_const: bool = False) -> str:
```
zhen8838: Why do you need to get a type str function to visit in it?

src/tilefoundry/inspection/python_printer.py
```diff
@@ -1292,23 +1284,27 @@ def _shape_tuple(shape: tuple) -> str:
...
+    """Enter the shared type visitor with the caller's block indentation."""
+    with _HIR_RENDERER.type_surface(indent=indent, const=is_const):
+        return _HIR_RENDERER.visit(value, ctx if ctx is not None else HirPrintContext())
+
+
 def _layout_str(layout: LayoutBase | None, indent: str = "") -> str:
```
zhen8838: What is the principle of this series of string functions? It shouldn‘t be written like this.

src/tilefoundry/inspection/python_printer.py
```diff
@@ -1329,17 +1325,50 @@ def _collect_all_meshes(
     return type_meshes, scope_meshes
 
 
-def _dedup_meshes(meshes: dict[int, Mesh]) -> dict[int, Mesh]:
-    """Collapse structurally identical descriptors before naming hoisted meshes."""
-    result: dict[int, Mesh] = {}
+def _mesh_name_map(meshes: dict[int, Mesh]) -> dict[int, str]:
```
zhen8838: Shouldn‘t this be recorded in the print context, and then there is an auxiliary function to construct Naming.

src/tilefoundry/inspection/python_printer.py
```diff
@@ -1329,17 +1325,50 @@ def _collect_all_meshes(
...
+    """
+    bound = {
+        names[identity]
+        for identity, mesh in meshes.items()
+        if mesh.names or identity in scope_mesh_ids
+    }
+    return {identity: name for identity, name in names.items() if name in bound}
+
+
 def _emit_header(
```
zhen8838: What does this emit header do?

src/tilefoundry/inspection/python_type_printer.py
```diff
@@ -0,0 +1,173 @@
...
+        return str(value)
+
+    def dtype_str(self, dtype: DType, ctx=None) -> str:
+        return dtype.name
+
+    def shape_tuple(self, shape: tuple, ctx=None) -> str:
+        values = tuple(self.dim_entry(entry, ctx) for entry in shape)
+        return f"({values[0]},)" if len(values) == 1 else "(" + ", ".join(values) + ")"
+
+    def shard_surface(self, value: ShardLayout, ctx=None) -> str | None:
```
zhen8838: Why does Python type print rely on Python print in reverse? This is the mistake. Obviously, you can generate it through Visitor. Why do you need to make an internal function?

tests/fixtures/tir/mma.py
```diff
@@ -11,8 +11,16 @@ class MmHandwritten:
     @prim_func(target=CudaTarget("nvidia.h200_sxm"))
     def mm_device(a: Tensor[(16, 16), "bf16"], b: Tensor[(16, 8), "bf16"], c: Tensor[(16, 8), "f32"]):
         with Mesh((Topology("thread", 32),), Layout((4, 8), (1, 4))) as _warp:
-            a_view = T.tensor_view(a, layout=ShardLayout(layout=Layout(shape=(2, 4, 2, 8, 2), strides=(1, 2, 8, 16, 128)), attrs=(S(1), S(3)), mesh=Mesh(topologies=(Topology(name="thread", size=32),), layout=Layout(shape=(4, 8), strides=(1, 4)), names=())))
-            b_view = T.tensor_view(b, layout=ShardLayout(layout=Layout(shape=(8, 2, 4, 2), strides=(1, 8, 16, 64)), attrs=(S(2), S(0)), mesh=Mesh(topologies=(Topology(name="thread", size=32),), layout=Layout(shape=(4, 8), strides=(1, 4)), names=())))
+            a_view = T.tensor_view(a, layout=ShardLayout(
```
zhen8838: The form of grammar sugar should be used here.

tests/fixtures/tir/sync.py
```diff
@@ -12,18 +12,25 @@ class SyncSquare:
...
-                    mesh=Mesh((Topology("thread", 128),), Layout((4, 32), (32, 1)), names=('w', 't')),
-                ), "rmem"])
+            view = T.tensor_view(a, layout=((4 @ m.w, 32 @ m.t), (32, 1)))
+            reg = T.alloc_tensor(tensor_type=Tensor[(4, 32), "f32", ((4 @ m.w, 32 @ m.t), (32, 1)), "rmem"])
             T.copy(view, reg)
             T.sync(m)
-            T.sync(Mesh(topologies=(Topology(name="thread", size=128),), layout=ComposedLayout(inner=None, offset=0, outer=Layout(shape=(1, 32), strides=(32, 1))), names=("w", "t")))
-            T.sync(Mesh(topologies=(Topology(name="thread", size=128),), layout=ComposedLayout(inner=None, offset=0, outer=Layout(shape=(2, 32), strides=(32, 1))), names=("w", "t")))
-            T.sync(Mesh(topologies=(Topology(name="thread", size=128),), layout=ComposedLayout(inner=None, offset=64, outer=Layout(shape=(2, 32), strides=(32, 1))), names=("w", "t")))
+            T.sync(Mesh((Topology("thread", 128),), ComposedLayout(
```
zhen8838: The form of grammar sugar should also be used here.

tests/fixtures/inspection/type_printer_sugar.analyzed.txt
```diff
@@ -0,0 +1,44 @@
+from __future__ import annotations
+
+from tilefoundry import func
+from tilefoundry.dsl import Tensor
+from tilefoundry.dsl.storage import gmem, rmem, smem
+from tilefoundry.dsl.tf import *  # noqa: F401, F403
+from tilefoundry.ir.types.shard import B, Layout, Mesh, P, S, ShardLayout, Topology
+
+@func(mesh=Mesh((Topology("cta", 4),), Layout((4,), (1,)), names=('tile',)))
```
zhen8838: This test case is unreasonable. First of all, there is no definition of the Topology of this func. Then why is the annotation of the following with mesh so complicated that it seems to be duplicated with the content in his body?

tests/fixtures/inspection/type_printer_sugar.printed.txt
```diff
@@ -0,0 +1,92 @@
...
+        mine = reshard(x, layout=(4 @ mesh.tile, 2, 16), storage=rmem)
+        v0 = reshard(mine, layout=Layout((8, 16), (16, 1)), storage=gmem)
+        escaped = reshard(x, layout=ShardLayout(
+            layout=Layout((2, 4, 16), None),
+            attrs=(S(0), B()),
+            mesh=Mesh((Topology("thread", 8),), Layout((2, 4), (4, 1)), names=('warp', 'lane')),
+        ), storage=rmem)
+        return (v0, held, frag, escaped)
+
+    @func(mesh=Mesh((Topology("cta", 4),), Layout((4,), (1,)), names=('tile',)))
```
zhen8838: There should be a function that defines Mesh((Topology("thread", 8),), Layout((2, 4), (4, 1)), names=('warp', 'lane')) on func, and then the acc of the parameter can be written in the form of syntax sugar.

tests/fixtures/inspection/type_printer_sugar.analyzed.txt
```diff
@@ -0,0 +1,39 @@
...
+        v6 = reshard(seed, layout=((2 @ mesh.warp, 4 @ mesh.lane, 2), (8, 2, 1)), storage=rmem)  # Tensor[(16,), "f32", ((2 @ mesh.warp, 4 @ mesh.lane, 2), (8, 2, 1)), "rmem"]; compute-cost; traffic traffic=gmem:r64/w0@cta:r64/w0,thread:r8/w0,rmem:r0/w64@cta:r0/w64,thread:r0/w8
+    folded = reshard(acc, layout=(2 @ mesh.warp, 4, 4 @ mesh.lane, 4), storage=rmem)  # Tensor[(8, 16), "f32", ((2 @ mesh.warp, 4, 4 @ mesh.lane, 4), (64, 16, 4, 1)), "rmem"]; compute-cost; traffic
+    summed = reshard(mixed, layout=((8, 16), {mesh.warp @ B()}), storage=rmem)  # Tensor[(8, 16), "f32", ((8, 16), (16, 1), {mesh.warp @ B()}), "rmem"]; compute-cost; traffic
+    for _ in range(3):  # Tuple[Tensor[(8, 16), "f32", ((2 @ mesh.warp, 4, 4 @ mesh.lane, 4), (64, 16, 4, 1)), "rmem"], Tensor[(8, 16), "f32", ((8, 16), (16, 1), {mesh.warp @ B()}), "rmem"]]; loop-footprint footprints=folded@rmem:8192/196608/24576,summed@rmem:131072/393216/393216 status=complete
```
zhen8838: For loop also does not need the print type here.

PR #170, opened 2026-09-20
fix(analysis): solve loop-aware buffer lifetimes

### LINKED ISSUE #169: fix(analysis): give a loop-carried value the residency the loop gives it
## Why

`#146` notes in passing that the lifetimes feeding `peak-footprint` are built on a
flattened value list carrying no loop nesting, and that this is a defect under either
reading of that number. That half has not landed; this issue states it on its own with
a repro, because an allocator is the consumer that needs it.

A value's residency is two indices into one flat, single-visit list
(`src/tilefoundry/analysis/memory.py:310-343`):

```python
index_by_id = {id(expr): index for index, expr in enumerate(values)}
last_by_id = dict(index_by_id)
for index, consumer in enumerate(values):
    for operand in expr_children(consumer):
        if id(operand) in last_by_id and index > last_by_id[id(operand)]:
            last_by_id[id(operand)] = index
```

The list is appended in visit order, and only `Call`/`Constant` land in it
(`memory.py:371`). A `LoopRegion` is visited (`memory.py:359-367`) but never appended,
and its `carried_args` are `Var`s, so they are not appended either. A value consumed
only through the carried edge therefore has no consumer in the list at all, and
`last_used_at` stays at its own index.

The peak sweeps those indices as if they were time (`memory.py:453-457`):

```python
for point in range(len(lifetimes) + 1):
    peak = max(peak, sum(item.bytes for item in rows
                         if item.defined_at <= point <= item.last_used_at))
```

### What it produces

A 256x256x256 bf16 GEMM tiled 64x64x32, with an f32 accumulator carried through the
`k` loop (program below, `analyze --memory --json`, at `e09a33e8`):

```text
binding   level  bytes   defined_at  last_used_at  persistent
v1:20     rmem   16384            3             3  False      <- acc = zeros([64,64] f32 rmem)
v9:24     rmem   16384           11            12  False
v10:24    rmem   16384           12            12  False
```

`v1:20` is the loop's carried-in accumulator. The loop body begins at index 4. The
report says its 16 KB is free from index 4 onward, and `persistent=False` keeps it out
of `persistent_bytes` too, so nothing in the report holds those bytes while the loop
runs.

The same flattening removes the rest of the loop structure: a value defined in a loop
body occupies one interval however many trips the loop takes, and two values in
different loops can never be shown to coexist even when they do.

## What

Build lifetimes on the `Scope` tree the memory family already walks, rather than on a
flat visit list:

- A value defined in a loop body is live for that scope's span, not for one index.
- A carried value is live from the loop's entry to its exit; the carried edge is a use.
- The peak sweep walks scope spans instead of list positions.

`analysis/scope.py` already carries what this needs — `Scope.domain`, `trips()`,
`is_variant`, `one_pass`, `reaching` — and `memory.py:359` already holds the child
scope for every `LoopRegion` it visits.

## Contract

`docs/spec/analysis.md` states `ValueLifetime` as `defined_at` / `last_used_at`
without saying what the two integers index. If they become scope-relative, that is the
place it gets written down.

## Risk

`peak-footprint` will rise for any program with a loop-carried value, which is most of
them, and `memory.py:468-475` refuses a program whose value exceeds a level's stated
capacity. A program that fits today on the strength of a point lifetime may start being
refused. That is the correction, but it moves report expectations and possibly some
programs from accepted to refused.

This is the reading `#146` is about. Filed separately because it is wrong under both of
that issue's two options, and because a buffer allocator cannot use the number at all
until it is fixed.

<details>
<summary>Repro program</summary>

```python
from tilefoundry import func, module
from tilefoundry.dsl import Mesh, Tensor, Topology, tf
from tilefoundry.target import CudaTarget

M = K = N = 256
BM = BN = 64
BK = 32


@module(entry="gemm", target=CudaTarget("nvidia.h200_sxm"),
        topologies=(Topology("cta", 1), Topology("thread", 128)))
class TILED:
    @func
    def gemm(a: Tensor[(M, K), "bf16"],
             b: Tensor[(K, N), "bf16"]) -> Tensor[(M, N), "bf16"]:
        with Mesh(("cta",), layout=(1,), names=("g",)) as cta:
            result = tf.zeros(Tensor[(M, N), "bf16"])
            for m in tile(M, BM):
                for n in tile(N, BN):
                    acc = tf.zeros(Tensor[(BM, BN), "f32", (BM, BN), "rmem"])
                    for k in tile(K, BK):
                        a_s = tf.reshard(a[m, k], (BM, BK), "smem")
                        b_s = tf.reshard(b[k, n], (BK, BN), "smem")
                        acc = acc + tf.reshard(
                            tf.matmul(tf.cast(a_s, dtype="f32"), tf.cast(b_s, dtype="f32")),
                            (BM, BN), "rmem")
                    result = tf.insert_slice(result, tf.cast(acc, dtype="bf16"), (m, n))
            return result
```

</details>

zhen8838:
## Why
- Fixes #169: loop-carried SSA values need physical reuse constraints, not summed logical lifetimes or unconditional carry exemptions.

## What
- Record structured SSA intervals plus real/synthetic use events, so carry ties reject later old-value uses without a second HIR traversal.
- Traverse HIR nodes with `AllocationConstraintVisitor`; derive exact loop-aware operand-to-result relations from the existing polyhedral access maps.
- Compile independent `insert_slice` result/destination and result/update relations only inside each gmem/smem address space.
- Keep every logical SSA box in CP-SAT, seed one complete feasible assignment, and return after the first validated placement.
- Report rmem as the largest logical value and lock all 52 existing Analysis corpus cases to exact per-level peaks.

## Contract
- gmem/smem `peak_bytes` is the returned placement high-water; rmem is the largest projected value.
- Capacity remains a post-solve, non-fatal `MemoryMetadata.errors`/JSON/`# error` report.
- `solver_status=feasible` does not claim a minimum.
- New liveness/allocation helpers remain package-internal.

## Risk
- Scheme B preserves iteration-dependent offsets in ISL but compiles translated subviews as containment. Exact affine byte offsets remain a localized future upgrade in `apply_constraints()`.


src/tilefoundry/analysis/liveness.py
```diff
@@ -0,0 +1,156 @@
...
+@dataclass(frozen=True)
+class Liveness:
+    """Definition-ordered intervals on one function-wide event timeline."""
+
+    intervals: tuple[LiveInterval, ...]
+    timeline_end: int
+
+
+@dataclass
+class _IntervalState:
```
zhen8838: This looks no different from LiveInterval.

zhen8838: Addressed in 63f33173: the visitor now stores `LiveInterval` directly and replaces the frozen value when extending a use, so the duplicate `_IntervalState` type is gone. The full source suite passes (1001 tests).

src/tilefoundry/analysis/allocation.py
```diff
@@ -0,0 +1,384 @@
...
+    """Union exact accesses to one buffer without discarding their maps."""
+    if not accesses or any(not access.exact for access in accesses):
+        return None
+    result = accesses[0].relation
+    for access in accesses[1:]:
+        result = result.union(access.relation)
+    return result.coalesce()
+
+
+def _proved_overlap_groups(
```
zhen8838: I think there may be something wrong with this algorithm.

src/tilefoundry/analysis/allocation.py
```diff
@@ -0,0 +1,384 @@
...
+
+
+def _lifetimes_overlap(left: AllocationValue, right: AllocationValue) -> bool:
+    """Whether two closed structured-SSA intervals share an event."""
+    return max(left.lifetime.defined_at, right.lifetime.defined_at) <= min(
+        left.lifetime.last_used_at, right.lifetime.last_used_at
+    )
+
+
+def _placement_hint(
```
zhen8838: What is this for?

zhen8838: Replaced in 840b9a26. `_placement_hint(groups, ...)` is gone. `_construct_feasible_seed()` consumes only BoolVars and box pairs already applied to the CP model; it chooses a mutually compatible set and supplies a complete address/order/peak hint. The hint exists only to reach the first feasible solution promptly, while CP-SAT still validates every box and constraint.

src/tilefoundry/analysis/allocation.py
zhen8838: Replaced in 840b9a26. The allocator no longer scans scopes into anonymous overlap groups or applies a global carried-value exemption. `AllocationConstraintVisitor` now visits each HIR node, derives one exact loop-aware `_OperandConstraint` per operand edge, proves liveness from explicit real/synthetic use events, and applies the relation only when both boxes belong to the same memory-level model. `InsertSlice` destination and update are independent edges.

Issue #173, opened 2026-09-21
feat(shard): let a layout carry a non-affine transform


zhen8838:
## Why

A WGMMA reads its A and B operands out of shared memory through a descriptor, and
the arrangement that descriptor names is a swizzled one on any kernel that goes fast.
`ir/types/shard/` cannot say it. `LayoutBase` has two subclasses: `Layout`, which is
shape and strides, and `ComposedLayout`, which is `inner(offset + outer(c))` over two
of those. Both are affine. A CUTLASS swizzle is an XOR permutation of the address,

```text
offset ^= ((offset >> s) & mask) << b
```

which no stride expresses. `grep -rn 'swizzle\|xor' src/tilefoundry/ir/types/shard/`
returns nothing.

The cost is measured, not estimated. A bf16 GEMM at M=8192 K=5120 N=17408 on an H200,
authored so that every arrangement in the emitted kernel is one the IR can state, and
bit-exact against `cublasLtMatmul`:

| | kernel-only | SM throughput | Mem Busy | DRAM | smem bank conflicts |
|---|---:|---:|---:|---:|---:|
| unswizzled, IR-stated | 7.216 ms | 21.4 % | 91.2 % | 6.4 % | 0 |
| same tiling, hand-swizzled | 3.551 ms | 50.7 % | 53.3 % | 15.3 % | — |
| cuBLASLt | 1.984 ms | 95.1 % | 80.9 % | 21.6 % | 4701 / 346117 |

Same tile, same instruction mix — the unswizzled kernel issues *fewer* instructions
(627M vs 690M) and has zero bank conflicts — and it is 2.03x slower by CUDA event,
2.34x by ncu duration. Memory is busy 91 % of the time while DRAM moves 6 %: the time
goes into the shared-memory path WGMMA reads its operands through.

So an author who stays inside the IR pays 2x, and an author who leaves it writes the
swizzle by hand and the program no longer says what it does.

## What

A `LayoutBase` that carries a non-affine transform, so a swizzled arrangement is a
layout like any other. Four things decide whether it is usable rather than merely
expressible:

1. **Composable into `ShardLayout`** — an operand pattern matches against
   `ShardLayout(layout=..., attrs=..., mesh=...)`, so the new kind has to sit in that
   `layout` slot.
2. **Evaluable, not only matchable** — given an index, it must give an offset. A
   consumer checking that a window and the buffer it is cut from place every element
   at the same address needs to *run* both layouts, which a pattern match does not
[...]

PR #174, opened 2026-09-21
feat(shard): express swizzled composed layouts as CuTe does

zhen8838:
## Why
- `Layout` and `ComposedLayout` could only state an affine coord-to-offset mapping, so the XOR swizzle a WGMMA shared-memory operand needs (issue #173) had nowhere to live.

## What
- `Swizzle(bits, base, shift)` ports CuTe's `Swizzle<B,M,S>` as a frozen functor that `ComposedLayout.inner` may carry. It is not a `LayoutBase`: its domain comes from the composition's `outer`.
- `layout_algebra` gains CuTe's swizzle specializations: `apply`, `cosize`, `coalesce`, the involution inverses, and `composition` — including `composition(Layout, Swizzle)`, which rebuilds the swizzle over the bits the layout leaves active so the answer is again a swizzle-inner composed layout rather than a `Swizzle` in `outer`.
- CUDA codegen renders the composition unchanged, `cute::make_composed_layout(cute::Swizzle<B,M,S>{}, cute::Int<offset>{}, ...)`, for both the layout type and the layout value.
- The device runtime reads steps off the affine part of a shard layout (CuTe deletes `stride()` on a `ComposedLayout`) and projects a composed one by folding the instance's origin into the composition's offset instead of advancing the engine pointer.
- One new test carries a `(128, 4)` f32 tile through swizzled SMEM: print/parse round-trip, codegen, nvcc, and the device, plus the emitted-layout assertion that running the kernel cannot make.

## Contract
- `ComposedLayout.inner` widens to `LayoutBase | Swizzle | None`; `outer`, `ShardLayout.layout` and `TensorType.layout` are unchanged, so a bare `Swizzle` is still not a tensor layout. `docs/spec/shard.md` §4 and the new §4.1 state it.
- The device projection rule for a composed shard layout is new public behavior: `docs/spec/runtime.md` §2.4.1 states that the origin reaches the composition's offset and not the engine pointer, and that a composed layout does not take the whole-tensor shortcut.
- A `Swizzle` states where an element lives, never which element a logical coordinate names: `shape`, `domain_rank`, the axis numbering a `Split` references and the access relation an op states are read off `outer` and are the unswizzled ones. `docs/spec/shard.md` §4.1 states it.
- A swizzled layout is still refused as a mesh execution scope, unchanged.

## Risk
- `Transpose`, `Reshape` and `Slice` carry the swizzle, since all three only relabel the domain and a window's start is a constant shift of the index that `offset` already holds. A register engine refuses it by name, as do a composed `inner` that is not a `Swizzle` and a `Swizzle` over a `ShardLayout` outer, which the shard slice path would drop.
- `composition(Layout, Swizzle)` and the inverses reject a non-zero composition offset: CuTe's rule for that case puts a `Swizzle` on the domain side, which this IR cannot represent.
- The ported `composition` and swizzled inverses have no caller in `src/` yet, so nothing exercises them; this PR ships them because the issue asks for CuTe's algebra, not because a code path needs them today.
- No WGMMA instruction and no performance gate ship here; this is the layout the operand needs, not the instruction that consumes it.


src/tilefoundry/ir/hir/tensor/slice.py
```diff
@@ -400,6 +400,13 @@ def _(call: "Call", ctx: "TypeInferContext") -> TensorType:
     layout_shape = shape
     source = x_ty.layout
     inherited_offset = 0
+    if isinstance(source, ComposedLayout) and source.inner is not None:
```
zhen8838: Why is this? Swizzle should not affect slice, right?

src/tilefoundry/ir/types/shard/layout.py
```diff
@@ -22,30 +22,93 @@ class Layout(LayoutBase):
...
+                f"bits={self.bits}"
+            )
+
+    @property
+    def bit_mask(self) -> int:
+        """CuTe ``bit_msk``: the ``bits`` low bits."""
+        return (1 << self.bits) - 1
+
+    @property
+    def yyy_mask(self) -> int:
```
zhen8838: Are you sure these three are from cute?

src/tilefoundry/ir/hir/tensor/slice.py
zhen8838: You are right -- carried instead of refused now.

A window only narrows the domain and shifts the index by a constant, and that constant is exactly what the composition's own offset states, so `Swizzle(off + start·stride + L_window(c))` is the same address the source gave that element. Slice now keeps the `Swizzle` the way Transpose and Reshape already kept it. Verified as `apply(sliced, c) == apply(source, start + c·step)` over a strided 2-D window.

Still refused: a composed `inner` that is not a `Swizzle`, and a `Swizzle` over a `ShardLayout` outer, because `_slice_shard_layout` rebuilds the composition with `inner=None` and would drop it.

src/tilefoundry/ir/types/shard/layout.py
zhen8838: Yes, all of it is `cute/swizzle.hpp`, same names.

The three checks above the anchor are its static_asserts (swizzle.hpp:61-63):
```cpp
static_assert(num_base >= 0,             "MBase must be positive.");
static_assert(num_bits >= 0,             "BBits must be positive.");
static_assert(abs(num_shft) >= num_bits, "abs(SShift) must be more than BBits.");
```
The three masks are swizzle.hpp:66-71, renamed only from `_msk` to `_mask`:
```cpp
using bit_msk = cute::constant<int, (1 << num_bits) - 1>;
using yyy_msk = cute::constant<int, bit_msk{} << (num_base + max(0,num_shft))>;
using zzz_msk = cute::constant<int, bit_msk{} << (num_base - min(0,num_shft))>;
static constexpr uint32_t swizzle_code = uint32_t(yyy_msk::value | zzz_msk::value);
```
`__call__` is its `apply`: `offset ^ shiftr(offset & yyy_msk, msk_sft)`. `zzz_mask` and `swizzle_code` are read by `composition(Layout, Swizzle)` and by the `make_swizzle<Y,Z>` check it ends in.

Issue #138, opened 2026-08-30
feat(evaluator): model the mesh so check can reference a coordinate-reading program


zhen8838:
## Why

`check` cannot establish a reference value for any program that reads its mesh
coordinate. The evaluator models a **single mesh participant** by design
(`docs/spec/evaluator.md:225`: `Reshard` "performs no cross-participant data
movement", `Local` "returns its operand's value for the single modelled
participant"), so `m.w` has no value:

```text
m.w desugars to:  Arange((4,)) -> Reshard(S(0)) -> Local() -> Reshape(())
intended:         [0,1,2,3]       split by mesh    this unit -> (1,)  -> scalar u
today:            [0,1,2,3]       identity         identity  -> (4,)  -> crash
```

```
$ tilefoundry check <fixture>:Strided --inputs random --out output --fn nan_inf
tilefoundry check: error: shape '[]' is invalid for input of size 4
```

This blocks the ordinary way to split a long reduction over a worker axis
(`b0 = t + m.w * BLK`) — the only placement with enough parallel units for
long-context attention. `tests/fixtures/placed/flash_split_k_decode.py:59`
already has this shape.

## What

Two facts worth recording before anyone picks this up.

**The unsound set is exactly one condition.** `parser/pattern_nodes.py:2903` is
the only site that constructs `Local`, so reading the coordinate is the only
source of per-unit variation. When nothing varies per unit, every unit computes
the same value and today's identity reading coincides with the truth — that is
why `Fixed`-shaped programs pass. A static marking

```text
unit_varying(e) = e is Local on a Split axis | any operand is unit_varying
```

measured over all of `tests/fixtures/` — 54 functions, 814 `Call` nodes — marks
**55 nodes (6.8%)**, all inside the two functions that read the coordinate.
Everything else is unaffected by whatever fix lands.

**Neither JAX nor Shardy has a reference evaluator to copy.** Shardy is
representation + propagation + SPMD partitioner; execution is delegated to the
backend. `shard_map`'s reference semantics is
`concatenate([f(blk) for blk in split(y)])` with **per-device local** shapes, and
that equation only holds for a collective-free body — ours has one
(`Reshard` S→B is an all-gather).

## Contract

`docs/spec/evaluator.md` §6 has to change: today's three statements
(single participant / `Reshard` does not move data / `Local` is identity) are
what any fix replaces. Until then, `check`'s answer for a mesh program is only
correct when no node reads the coordinate.

## Risk

[...]

PR #175, opened 2026-09-21
feat(analysis): support unit-dependent loop starts

zhen8838:
## Why
- Persistent kernels need loop starts and stops derived from mesh coordinates so each unit can own a distinct tile range without losing analysis precision.
- Existing loop domains treated every non-literal bound as runtime data and could not preserve bounded unit parameters through access analysis.

## What
- Carry bounded mesh coordinates through dimension ranges, loop domains, trip counts, access relations, and allocation proofs.
- Support three-argument `tile(start, stop, step)` syntax, printing, and round trips.
- Add realistic tiled and grid-stride persistent GEMM fixtures covering all four analysis families, unit partitioning, precision, memory peaks, and reference tiling arithmetic.

## Contract
- `tile()` accepts two arguments or three arguments; the three-argument form matches `range(start, stop, step)` semantics.
- Loop `start` and `extent` may depend on bounded unit values. `step` must remain literal because parametric strides have no Presburger representation.
- Type-inference rules may return `TypeInferResults` with a value range; returning a bare `Type` remains valid.
- `RangeMetadata` follows the `expr.type` lifecycle. Inference that does not write types also does not write ranges.
- `dim_range` returns `None` when no bound is available; unsupported symbolic divisors still raise.
- `Access.exact` is replaced by `AccessPrecision` with `EXACT`, `WIDENED`, and `UNKNOWN`. Relations with bounded parameters may remain exact.
- Parameterized cardinality enumerates boxes up to `PARAM_POINT_LIMIT` for the true maximum and returns `None` above the limit.
- Public analysis helpers now include `range_expr`, `count`, `param_points`, `has_unbounded_param`, `PARAM_POINT_LIMIT`, `UnboundedParameterBox`, and `ParameterBoxTooLarge`.
- The analysis preflight does not write `Expr.type`; the previous specification wording claiming it did was corrected.

## Risk
- The flat fixture remains `WIDENED`: affine loop binding does not support floor-div/mod.
- Mesh axis naming and x/y/z fallback logic remain duplicated between parser and printer and should move to a shared mesh helper.
- Loop bounds still lack ShapeDim validation, so `normalize_dim` may retain an invalid static object.
- Parameter boxes above 4096 points make peaks/trips unknown.
- Converters are not included in whole-function range reinference, matching prior behavior; no range consumer currently reads them.


docs/spec/analysis.md
```diff
@@ -1218,6 +1223,20 @@ call site, source expressions shared by identity remain one shared expression
 in the clone; sharing never aliases the independently cloned body of another
 call site.
 
+A loop `start` or `extent` MAY depend on the unit running it when every runtime
```
zhen8838: Don't say such a prose.

src/tilefoundry/analysis/allocation.py
```diff
@@ -24,7 +24,7 @@
 from .errors import AnalysisError
 from .liveness import Liveness
 from .metadata import ValueLifetime
-from .scope import Access, Scope
+from .scope import Access, AccessPrecision, Scope
```
zhen8838: This member should not be put in the scope.

src/tilefoundry/analysis/scope.py
```diff
@@ -14,11 +15,18 @@
 from tilefoundry.ir.hir.tensor.reshape import Reshape
 from tilefoundry.ir.hir.tensor.slice import Slice
 from tilefoundry.ir.types import TensorType
-from tilefoundry.ir.types.dim import DimVar
+from tilefoundry.ir.types.dim_isl import range_expr
```
zhen8838: What is RangeExpr?

src/tilefoundry/analysis/scope.py
```diff
@@ -14,11 +15,18 @@
...
 from tilefoundry.ir.types.shape_helpers import static_dim_value
 from tilefoundry.ir.types.utils import local_type_of
 from tilefoundry.ir.visitor import expr_children
-from tilefoundry.utils.isl_utils import cardinality
+from tilefoundry.utils.isl_utils import (
+    PARAM_POINT_LIMIT,
+    ParameterBoxTooLarge,
+    UnboundedParameterBox,
+    count,
```
zhen8838: What is the count here? Didn't we have cardinality?

src/tilefoundry/analysis/scope.py
```diff
@@ -222,6 +276,48 @@ def _induction_of(loop: LoopRegion) -> str:
...
+            f"literal, or state it as an open dimension"
+        )
+    raise AnalysisError(
+        f"loop {_induction_of(loop)!r} takes its {which} from "
+        f"{value_label(value) or 'a value'!r}, which the program computes at run time; "
+        f"analysis needs a literal {which} or a stated value range"
+    )
+
+
+def _render_bound(
```
zhen8838: The Scope file bears too many unnecessary responsibilities, and the content here should be split again.

src/tilefoundry/analysis/scope.py
```diff
@@ -455,4 +576,11 @@ def walk_scopes(root: Scope) -> Iterator[Scope]:
         yield from walk_scopes(child)
 
 
-__all__ = ["Access", "Scope", "ScopeBuilder", "build_scopes", "walk_scopes"]
+__all__ = [
+    "Access",
+    "AccessPrecision",
+    "Scope",
```
zhen8838: The name Scope also needs to be renamed. It's hard to understand what it does.

src/tilefoundry/inspection/printer_base.py
```diff
@@ -136,6 +140,28 @@ def visit_Call(self, value: Call, ctx=None) -> str:
                 return f"{name}({args})"
         return self.visit_program_call(value, ctx)
 
+    @staticmethod
+    def _mesh_coordinate_text(value: Call, target: MeshCoord, ctx) -> str:
```
zhen8838: Why does this function need to be static?

src/tilefoundry/inspection/python_printer.py
```diff
@@ -856,7 +855,11 @@ def _emit_loop_region(region: LoopRegion, level: str) -> None:
         start = printer.visit(region.start, ctx)
```
zhen8838: Why does the printing here need to be judged? Can't you just default to 3 parameters?

src/tilefoundry/ir/types/dim_isl.py
```diff
@@ -309,20 +318,26 @@ def normalize_dim_entries(value):
     return value
 
 
-def dim_range(dim) -> tuple[int, int]:
+def dim_range(dim) -> tuple[int, int] | None:
```
zhen8838: I need to understand what the whole document is doing.

src/tilefoundry/parser/ast_pattern.py
zhen8838: Why do you need to get a new namespace? I don't think it's necessary.

src/tilefoundry/parser/ast_pattern.py
```diff
@@ -1547,6 +1555,15 @@ def _resolve_reference(node: ast.AST, context: MatchContext) -> object:
         raise ParseError.from_node(node, context, f"undefined static name {node.id!r}")
     if isinstance(node, ast.Attribute):
         owner = _resolve_reference(node.value, context)
+        if isinstance(owner, Mesh) and not hasattr(owner, node.attr):
```
zhen8838: You shouldn't judge the name here. It's a very dangerous thing.

src/tilefoundry/parser/pattern_nodes.py
```diff
@@ -1614,6 +1615,20 @@ def construct(match, children, context):
         elif match.branch_id == "static_attribute":
             owner = children["owner"]
             attribute = match.captures["attribute"]
+            if isinstance(owner, runtime.Mesh) and not hasattr(owner, attribute):
```
zhen8838: We shouldn't do this here either.

src/tilefoundry/utils/isl_utils.py
```diff
@@ -2,36 +2,137 @@
 
 from __future__ import annotations
 
+from itertools import product
+
 import isl
 
-__all__ = ["as_multi_aff", "cardinality", "equates", "involved_dims"]
+__all__ = [
```
zhen8838: Too much content has been added here at once, and it feels repetitive. Especially count and cardinality

src/tilefoundry/visitor_registry/visitors.py
```diff
@@ -56,11 +64,14 @@ class TypeInferVisitor(ExprVisitor[Type]):
     stale ``expr.type``.
     """
 
-    def __init__(self, *, memo=None, owns_body: bool = True) -> None:
+    def __init__(
```
zhen8838: Why is the type inference visitor not in a separate file, but in this public file?

tests/analysis/test_analysis_families.py
```diff
@@ -163,6 +166,22 @@ def _calls(function) -> tuple[Call, ...]:
     return tuple(expr for expr in collect_exprs(function.body) if isinstance(expr, Call))
 
 
+def test_one_pass_maximizes_over_bounded_mesh_parameters() -> None:
```
zhen8838: This test should not be added.

tests/fixtures/placed/persistent_gemm_flat.py
```diff
@@ -0,0 +1,70 @@
...
+                    acc = acc + tf.reshard(product, (BM, BN), "rmem")
+                out = tf.insert_slice(
+                    out,
+                    tf.reshard(acc, (BM, BN), "gmem"),
+                    (mi, ni),
+                )
+            return out
+
+
+persistent_gemm_flat = PersistentGemmFlat.entry_function()
```
zhen8838: Why did you export this?

tests/integration/test_persistent_gemm.py
```diff
@@ -0,0 +1,147 @@
...
+            acc = torch.zeros(flat.BM, flat.BN, dtype=torch.float32)
+            for ki in range(0, flat.K, flat.BK):
+                acc += a[mi : mi + flat.BM, ki : ki + flat.BK].float() @ b[
+                    ki : ki + flat.BK, ni : ni + flat.BN
+                ].float()
+            result[mi : mi + flat.BM, ni : ni + flat.BN] = acc
+    return result
+
+
+def test_reference_tiling_index_arithmetic_matches_dense_gemm() -> None:
```
zhen8838: Delete all this test file. You should measure it directly through the test fixture. End-to-end analysis test and evaluator test.

tests/ir/test_loop_region_range_nested.py
```diff
@@ -72,6 +80,16 @@ def test_range_start_step():
     assert torch.allclose(out.reshape(()), x[1:n:2].sum()), (n, out)
 
 
+def test_tile_start_stop_step_maps_to_loop_region_fields() -> None:
```
zhen8838: Delete this test.

tests/parser/test_mesh_visibility.py
```diff
@@ -105,3 +107,23 @@ def test_region_boundaries_capture_external_regions_through_args() -> None:
 
     import_dsl(_diagnostic("region_rebind"), "RegionRebind")
     import_dsl(_diagnostic("region_tuple_rebind"), "RegionTupleRebind")
+
+
+def test_three_argument_tile_accepts_mesh_coordinate_bounds() -> None:
```
zhen8838: Delete this test.

docs/spec/analysis.md
zhen8838: Addressed in 9bc10d16: the added prose is now expressed as structured `- constraints:` contract entries.

src/tilefoundry/analysis/allocation.py
zhen8838: Addressed in daa33960 and 279643b5: `Access` and `AccessPrecision` now live in `analysis/access.py`, and allocation imports them from their owner.

src/tilefoundry/analysis/scope.py
zhen8838: Addressed in 2a657b56: `range_expr` is now `dim_to_isl_expr`, naming both the input concept and the conversion direction.

src/tilefoundry/analysis/scope.py
zhen8838: Addressed in 890e9c14 and 49319704: `cardinality` is the sole public counting entry, `_count` is private, and the empty-parameter-context result is explicitly tested as zero.

src/tilefoundry/analysis/scope.py
zhen8838: Addressed in daa33960 and 279643b5: the old scope module is split into `iteration_scope.py`, `access.py`, and `loop_domain.py`. Parameter elimination keeps explicit inputs because those values have no shared lifecycle; wrapping them would hide rather than reduce responsibility.

src/tilefoundry/analysis/scope.py
zhen8838: Addressed in daa33960: `Scope` is now `IterationScope`, which states that the tree represents function and loop iteration domains.

src/tilefoundry/inspection/printer_base.py
zhen8838: Addressed in 66cd89c3: `_mesh_coordinate_text` is now an instance method.

src/tilefoundry/inspection/python_printer.py
zhen8838: Kept the two-argument spelling when `start == 0` by design. Always printing three arguments would churn eight fixture outputs and two tutorial goldens and make every existing zero-start loop more verbose; the third argument is emitted when a nonzero start requires it.

src/tilefoundry/ir/types/dim_isl.py
zhen8838: Addressed in 2a657b56: `dim.py` and `dim_isl.py` now have module-level ownership documentation, and the public conversion names state their direction (`dim_to_isl_expr`, `shape_to_isl_domain`, and `isl_to_dim`).

src/tilefoundry/parser/ast_pattern.py
zhen8838: Addressed in 66cd89c3: `inference_type` is no longer injected into the parser runtime namespace; the parser imports it directly from its owning module.

src/tilefoundry/parser/ast_pattern.py
zhen8838: Addressed in 66cd89c3: the Mesh attribute-name guard was removed from `_resolve_reference`. Consolidating mesh axis naming and x/y/z fallback across parser and printer remains a separate follow-up and is called out in the PR Risk section.

src/tilefoundry/parser/pattern_nodes.py
zhen8838: Addressed in 66cd89c3: the duplicate Mesh attribute-name guard was removed from `StaticReferencePattern.construct`; this layer no longer decides whether the attribute name is valid.

src/tilefoundry/utils/isl_utils.py
zhen8838: Addressed in 890e9c14 and 49319704: public `count` was removed and `cardinality` owns point counting, with `_count` only an implementation detail. Parameter-box helpers remain in `utils/isl_utils.py` because they are IR-independent and preserve the existing IR-to-utils dependency direction.

src/tilefoundry/visitor_registry/visitors.py
zhen8838: Addressed in bedef606: `TypeInferVisitor` and `inference_type` now live in `visitor_registry/typeinfer.py`; `visitors.py` retains only verify, codegen, and cost visitors. `TypeInferResults` keeps its name because it carries both a type and an optional value range, and renaming that public carrier is unrelated churn.

tests/analysis/test_analysis_families.py
zhen8838: Addressed in 94681a93: this dedicated one-pass test was removed. The persistent schedule expectations now ride the existing placed-program inventory in de334e5f.

tests/fixtures/placed/persistent_gemm_flat.py
zhen8838: Addressed in 94681a93: the unused entry-function exports were removed from both persistent fixtures.

tests/integration/test_persistent_gemm.py
zhen8838: Addressed in 94681a93 and de334e5f: the integration test file was deleted, and analysis coverage moved into the existing fixture-driven placed-program inventory. Numerical per-device evaluation remains deferred to issue #138 because the evaluator intentionally has no current MeshCoord participant.

tests/ir/test_loop_region_range_nested.py
zhen8838: Addressed in 94681a93: this dedicated loop-region range test was removed as requested.

tests/parser/test_mesh_visibility.py
zhen8838: Addressed in 94681a93: the dedicated three-argument tile and axis-name parser coverage was removed as requested.

docs/spec/code-organization.md
```diff
@@ -34,8 +34,10 @@ truth for the directory's structure and invariants.
...
 | `analysis/errors.py` | [analysis](./analysis.md) | `AnalysisError`, the one diagnostic the whole analysis layer raises, so catching an analysis failure catches every analysis failure rather than the subset the caller happened to import. |
-| `analysis/visitor.py` | [analysis](./analysis.md) | The per-call `AnalyzeContext`, carrying the shared root/current lexical `Scope` while a family traverses its work. |
-| `analysis/scope.py` | [analysis](./analysis.md) | The shared `Scope` tree and `Access` relations built once from normalized HIR; families query these views instead of constructing parallel structure. |
+| `analysis/visitor.py` | [analysis](./analysis.md) | The per-call `AnalyzeContext`, carrying the shared root/current `IterationScope` while a family traverses its work. |
+| `analysis/iteration_scope.py` | [analysis](./analysis.md) | The shared `IterationScope` tree built once from normalized HIR; families query it instead of constructing parallel structure. |
+| `analysis/access.py` | [analysis](./analysis.md) | Access relations resolved against the authored iteration scopes. |
```
zhen8838: Why isn't access put in affine.py?

src/tilefoundry/analysis/footprint.py
```diff
@@ -7,8 +7,8 @@
 import isl
 
 from tilefoundry.ir.types import TensorType
+from tilefoundry.ir.types.dim_isl import index_set
```
zhen8838: Index_set should not come from such a dim_isl

src/tilefoundry/ir/types/dim_isl.py
```diff
@@ -343,31 +363,46 @@ def bind(name: str, dim, lo: int, hi: int) -> None:
...
+        if (bound := bounds[name]) is not None
+    ]
     prefix = f"[{', '.join(names)}] -> " if names else ""
     if not dims:
         return isl.set(prefix + "{ [] }"), param_map
     body = f"{{ [{', '.join(dims)}] : {' and '.join(constraints)} }}"
     return isl.set(prefix + body), param_map
 
 
+def index_set(shape: tuple) -> isl.set | None:
```
zhen8838: This function is related to shape and should not be put in dim isl.

docs/spec/code-organization.md
zhen8838: Addressed in 1942765d (M6). affine.py was renamed to loop_terms.py and now only converts HIR into LoopTerm; it contains no isl usage. Access stays in analysis/access.py because it assembles isl relations, and widest_allowed moved there as well. The layer boundary is whether the code touches isl.

src/tilefoundry/analysis/footprint.py
zhen8838: Addressed in f18b3f4b (M5). index_set moved together with shape_to_isl_domain into ir/isl_interop.py. The old ir/types/dim_isl.py module was removed; analysis no longer obtains this shape/isl bridge from a dim-types module.

src/tilefoundry/ir/isl_interop.py
```diff
@@ -343,31 +363,46 @@ def bind(name: str, dim, lo: int, hi: int) -> None:
...
+        if (bound := bounds[name]) is not None
+    ]
     prefix = f"[{', '.join(names)}] -> " if names else ""
     if not dims:
         return isl.set(prefix + "{ [] }"), param_map
     body = f"{{ [{', '.join(dims)}] : {' and '.join(constraints)} }}"
     return isl.set(prefix + body), param_map
 
 
+def index_set(shape: tuple) -> isl.set | None:
```
zhen8838: Addressed in f18b3f4b (M5). index_set and shape_to_isl_domain now live together in ir/isl_interop.py, which owns dim/shape and isl interoperation. ir/types/dim_isl.py no longer exists, while ir/types/dim.py is limited to definitions and constructors.

